PR #6, opened 2016-12-05
add --save-dev [skip ci]

hzoo:
We usually add plugins as devDeps

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Full, scoped and component-friendly CSS support for JSX (SSR + browser).
 Firstly, install the package:
 
 ```bash
-$ npm install --save styled-jsx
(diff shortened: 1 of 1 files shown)


leo:
Woopsie! You're right, of course. Thanks! ❤️

nkzawa:
I think some files are for runtime, so it has to be `--save` ?

leo:
The client, yep. Fixing!

hzoo:
Oh you mean `import _jsxStyleInject from 'styled-jsx/inject'`

hzoo:
It was just weird to me about babel-core being a dep ref https://github.com/zeit/styled-jsx/pull/7

PR #7, opened 2016-12-05
remove unused babel-cli, move babel-core to devDep

hzoo:
The `babel` npm package currently doesn't do anything (you usually mean to install babel-cli).

`babel-core` is only used in the tests https://github.com/zeit/styled-jsx/blob/master/test/index.js#L5 so can be a devDep as well

And I didn't see babel-cli referenced anywhere?

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,13 +4,11 @@
   "license": "MIT",
   "main": "./flush.js",
   "dependencies": {
-    "babel": "^6.5.2",
-    "babel-core": "^6.18.2",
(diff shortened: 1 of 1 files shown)


hzoo:
Actually.. I think babel-cli should be there but there's no script to build the files?

hzoo:
Was it intentional to not compile the files https://unpkg.com/styled-jsx@0.0.1?

leo:
#general in https://zeit.chat

PR #9, opened 2016-12-05
Adding repository reference

TehShrike:


leo:
Thanks! But please let's go with `zeit/styled-jsx` as the prop value.

PR #35, opened 2016-12-11
Improved hashing performance with faster algorithm

rtsao:
[ESBench](https://esbench.com/bench/584ddcf9d5686f990078bac8)

I use a JS-optimized djb2 hash implementation in [CSJS](https://github.com/rtsao/csjs) since it's faster and more lightweight than murmurhash2 (which is important in that use case because it actually runs at runtime). This obviously doesn't matter as much for styled-jsx since it's build-time.


lib/djb2.js
```diff
@@ -0,0 +1,12 @@
+// djb2 via https://github.com/darkskyapp/string-hash
```
nkzawa: It looks we can just use `string-hash` ?

rauchg:
Merged. Thank you!

PR #46, opened 2016-12-19
Better SSR methods

### LINKED ISSUE #22: Same styles are applied twice when Server-Side Rendering
It seems we have to skip generating style tags on client if it's already rendered on SSR.

rauchg:
Fixes #22 

- Adds `styled-jsx/server`
  - `flush()` (`default`) returns an array of `React.Element` for each `<style>`
  - `flushToHTML` returns an HTML string

Additionally

- Both new methods include an `id` in the resulting `<style>` that the client rendering part can use to de-dupe. 
- The de-duping logic only checks the DOM once and works transparently, without manual initialization by the user

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,12 +150,12 @@ export default () => (
 
 The main export flushes your styles to an array of `React.Element`:
 
-```
+```js
 import React from 'react'
 import ReactDOM from 'react-dom/server'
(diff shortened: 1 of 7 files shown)


nkzawa:
It looks we have to fix styles (CI is failing)

Issue #57, opened 2016-12-25
CSS compiler issues: comments and commas


rauchg:
The following doesn't compile well, unless I take out the comment

```css
.story {
  /* todo: probably a more flex-y way of doing this? */
  flex: 100;
  display: inline-block;
}
```

And based on how stylis.js was originally written, it uses `split` with `,` for selectors, where it should actually keep track that the `,` is not inside a selector for example., so this will break:

```css
.story[title=","] {
  display: inline-block;
}
```

giuseppeg:
@rauchg fixing the comments should be trivial, do you want to preserve or strip them out?
The second case is tricker. Does it break when there are properties with multiple values separated by comma like `box-shadow: 0 0 10px black, inset 0 0 5px black` ?

rauchg:
- I think we should strip / minify as much as possible, since this is the _compilation_ step
- I mean for selectors. Reading the code I realized [this](https://github.com/zeit/styled-jsx/blob/master/lib/style-transform.js#L197) is too naive, and @thysultan has already addressed it upstream if you need inspiration

thysultan:
originally comments where not handled property there where edge cases that escaped, though upstream they are handled well.

giuseppeg:
Adding one more here – the `:root` selector shouldn't be namespaced cc @nkzawa

rauchg:
+1

giuseppeg:
fixed by #134

PR #58, opened 2016-12-25
CSS Compiler: Strip out comments

giuseppeg:
See #57


lib/style-transform.js
```diff
@@ -31,6 +31,7 @@
     var regPrefix    = /((?:transform|appearance):.*?;)/g;
     var regSpaces    = /  +/g;
     var regAnimation = /(,|:) +/g;
+    var regComments = /\/\*[^]*?\*\//g;
```
thysultan: @giuseppeg You may also want to handle inline js comments `// comment`

giuseppeg: I didn't do that on purpose but I can strip those too cc @rauchg

PR #59, opened 2016-12-26
Fix flush, remove advanced APIs

nkzawa:
Fixes https://github.com/zeit/next.js/issues/496

Additionally I removed advanced APIs (`styled-jsx/flush` and `styled-jsx/memory`) since we alternatively have `styled-jsx/server`.


src/style.js
```diff
@@ -23,6 +23,18 @@ export default class extends Component {
...
 }
 
+export function flush() {
+  const ret = {}
+
+  for (const {props} of components) {
+    ret[props.styleId] = props.css
+  }
+
+  components = []
```
nkzawa: We have to reset `components` :(

PR #61, opened 2016-12-29
Add support for expressions within template literals

### LINKED ISSUE #25: Expressions within template literal don't work
I would understand if dynamic properties didn't work, but dynamic values don't seem to be working either. Compiling succeeds, but the pieces with the expression seem to be omitted from the inserted script tag.

```
<div>
	<div id={this.state.calendarId}></div>
	<style jsx>{`
			#${this.state.calendarId} .fc-button.fc-state-default {
				border: 2px solid ${buttonColor};
			}
	`}</style>
</div>
```
image

---

```
const buttonColor = 'red';

return (
	<div>
		<div id="foo"></div>
		<style jsx>{`
			#foo .fc-button.fc-state-default {
				border: 2px solid ${buttonColor};
			}
		`}</style>
	</div>
);
```
image

---

```
<div>
	<div id="foo"></div>
	<style jsx>{`
		#foo .fc-button.fc-state-default {
			border: 2px solid red;
		}
	`}</style>
</div>
```
image

Thank you!

giuseppeg:
Fixes #25


giuseppeg:
fyi this patch doesn't allow expressions in the selector name like `.col-${xsmall}` not sure if that's necessary or that useful.

sarukuku:
Waiting for this to land. 🚀

giuseppeg:
@rauchg, @nkzawa I think that this branch is ready for review.
Let me know if you need context but hopefully this solution makes sense and it is not too dumb :)

src/babel.js
```diff
@@ -17,23 +18,98 @@ const STYLE_COMPONENT_CSS = 'css'
 export default function ({types: t}) {
   const findStyles = children => (
     children.filter(el => (
```
timneutkens: Just a small thing, think you can destructure this param `({node}) => (` which allows you to remove `el` in front of `node`

giuseppeg: oh right good catch!

timneutkens: :fire:

rauchg:
Are we erroring upon usage of vars from the closure?

giuseppeg:
No but we could although it might add some overhead or be not that straightforward if we keep allowing things like call expressions

giuseppeg:
basically you want to make sure that people don't use `props` right?

rauchg:
Correct. If this doesn't fail it will create extremely difficult-to-understand bugs to the uninitiated:

```jsx
export default ({ some, props}) => {
  <div>
    <style jsx>{`
      div { color: ${some} }
    `}</style>
  </div>
}
```

giuseppeg:
@rauchg any identifier in the scope or just props?

```jsx
export default () => {
  const some = 'red' // allowed or not?

  return (<div>
    <style jsx>{`
      div { color: ${some} }
    `}</style>
  </div>)
}
```

rauchg:
I'd block anything from the scope. After all, if it's a constant, you should probably syntactically move it outside the scope…

src/babel.js
```diff
@@ -16,24 +17,120 @@ const STYLE_COMPONENT_CSS = 'css'
...
+  }
 
-  const makeStyledJsxTag = (id, transformedCss) => (
-    t.JSXElement(
+  const makeStyledJsxTag = (id, transformedCss, isTemplateLiteral) => {
+    let css
+    if (isTemplateLiteral) {
+      // build the expression from transformedCss
+      traverse(
+        parse(`\`${transformedCss}\``).ast,
```
nkzawa: I think we should use https://github.com/babel/babylon

giuseppeg: babel-core's transform uses babylon behind the wood

nkzawa: yep, but I meant it would be better to use the parser anyway.

sedubois: Behind the wood 😄👍

giuseppeg: rofl typo. @nkzawa yea sorry I was rushing. In fact I thought of doing that initially but then I didn't feel like adding the extra dependency but if you think that it is preferable I have no problem with that.

rauchg:
Needs rebase

giuseppeg:
@rauchg you got it

sarukuku:
Any updates on this? 🚀

giuseppeg:
@sarukuku I haven't had a lot of free time recently. Anyway we want to benchmark this before merging – I will try to push a minimal benchmark today after work.

giuseppeg:
301 to #80 :)

PR #63, opened 2016-12-30
Switch from rAF to Promise to fix FOUC in Safari

### LINKED ISSUE #52: FOUC in Safari
After converting a site to `next@beta` and `styled-jsx`, I started to notice a FOUC in Safari when I transitioned pages.

Having already looked at `styled-jsx` my hunch was the rAF callback that updates the styles in the DOM is called in a different order in Safari than Chrome. I set a break-point within the rAF callback in both Safari and Chrome and sure enough the order of operations is different.

**Chrome:**
1. unmount outgoing `<style jsx>` components
2. render incoming `<style jsx>` components
3. mount incoming `<style jsx>` components
4. __rAF callback fires, update `<style>`s in `<head>`__
5. __paint__

**Safari:**
1. unmount outgoing `<style jsx>` components
2. render incoming `<style jsx>` components
3. mount incoming `<style jsx>` components
4. __paint__
5. __rAF callback fires, update `<style>`s in `<head>`__

Safari painted the updated DOM tree *before* `styled-jsx`'s rAF callback was called. While in Chrome, the rAF callback is called first, then Chrome paints. This is why I'm seeing FOUC in Safari but not in Chrome.

I didn't have time to dig into a possible solution for this, but wanted to bring it up in case this wasn't already a known issue.

ericf:
This fixes the FOUC issue in Safari by switching from `requestAnimationFrame()` to `Promise#then()` for the debouncing/batching mechanism. `requestAnimationFrame()` causes FOUC in Safari, `setTimeout()` causes FOUC in Chrome, while `Promise#then()` ensures micro task enqueueing of styles update before paint in Safari, Chrome, and Firefox.

This maintains the same semantics where only the latest call to `updateOnClient()` causes the rendering of styles by "cancelling" the previously created promises by making them noops.

Fixes #52
/cc @rauchg @nkzawa

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -3,7 +3,7 @@ import render from './render'
 
 const update = typeof window === 'undefined' ? doRender : updateOnClient
 let components = []
-let requestId
+let updatePromise
 
 export default class extends Component {
   componentWillMount() {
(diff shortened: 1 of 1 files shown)


nkzawa:
This is awesome ! I didn't notice this solution, thanks 👍

PR #75, opened 2017-01-15
<style> as a top-level component

### LINKED ISSUE #72: Make it possible to have <style> as a top-level component
Hello!

I had this scenario today: https://gist.github.com/rickyrauch/a9aa62854d2ebf9bed22d4bd14561de8

I fixed this wrap it in a `<div>`

Thanks

giuseppeg:
Fixes #72


src/babel.js
```diff
@@ -15,16 +15,30 @@ const STYLE_COMPONENT_ID = 'styleId'
...
+    const firstChild = children[0]
+    if (children.length === 1 && !isStyledJsx(firstChild)) {
+      const parent = firstChild.parentPath
+      return (
+        isStyledJsx(parent) && isGlobalEl(parent.node.openingElement) ?
+        [parent.node] : []
+      )
+    }
+    return children.filter(isStyledJsx).map(({node}) => node)
+  }
```
nkzawa: I wonder if we can't traverse from parent node instead of `findStyles(path.node.children)` ? It might makes things simpler.

giuseppeg: oh silly me, good idea.

nkzawa:
👍 👍 👍

PR #78, opened 2017-01-16
patch animation namespacing

thysultan:


nkzawa:
wow, thanks for your response 😄

PR #79, opened 2017-01-16
Benchmark Babel transform

giuseppeg:
```
Babel transform x 104 ops/sec ±1.32% (562 runs sampled)
```


benchmark/babel.js
```diff
@@ -0,0 +1,30 @@
+import path from 'path'
+import Benchmark from 'benchmark'
+import {transformFile} from 'babel-core'
+
+import plugin from '../src/babel'
+
+const transform = (file, opts = {}) => (
+  new Promise((resolve, reject) => {
+    transformFile(path.resolve(__dirname, file), {
```
rauchg: shouldn't we avoid the benchmark doing I/O?

giuseppeg: Yeah but I am not sure if `benchmark` supports async setup https://github.com/bestiejs/benchmark.js/pull/174

I can read the file synchronously before exporting the benchmark wdyt?

giuseppeg:
`155 ops/sec ±1.39% (573 runs sampled)`

PR #97, opened 2017-01-27
Fix multiple animations transpilation

nkzawa:
This PR fixes the incorrect transpilation of multiple animations definition.


lib/style-transform.js
```diff
@@ -184,6 +184,7 @@
 
                             // animation name is anything not in this list
                             if (
+                              prop &&
```
giuseppeg: you could as well move `isNaN(parseInt(prop))` here I suppose

nkzawa: I wonder why. For performance ?

giuseppeg: nvm ignore this comment, I overlooked this I thought that `isNaN(parseInt(prop))` was negated `! ...`.

Issue #95, opened 2017-01-25
Multiple selectors with global prepends unexpectedly


cgood92:
I have written the following rule:

```css
nav :global(a.active), nav :global(a:hover) {
   ...
}
```

The built/compiled version of this css piece results in this unexpected result:

```css
nav[data-jsx="3533501187"] a.active[data-jsx="3533501187"], nav[data-jsx="3533501187"] a:hover {
   ...
}
```

The code block above prepends the first `:global(a.active)` with a jsx id for some reason (the `[data-jsx="3533501187"]` part), though it shouldn't because it is in the `:global()`.  The second one works correctly, as I would expect (where it is not prepended with a jsx qualifier).  

I expect the result to look like this:

```css
nav[data-jsx="3533501187"] a.active, nav[data-jsx="3533501187"] a:hover {
   ...
}
```

Where the selector ends at `a.active`, not `a.active[data-jsx="3533501187"]`.

What is going on here?  Whatever selector first, it prepends with a `[data-jsx]`.  If I swap them around, it still prepends the first of the comma-separated selectors.

I've tried the following things as well (to no success):

```css
/* does not work */
nav :global(a.active, a:hover) {
   ...
}

/* does not work */
nav :global(a.active), :global(a:hover) {
   ...
}
```

PR #98, opened 2017-01-27
Fix global selectors transform

nkzawa:


giuseppeg:
@nkzawa which bug is this branch fixing? Can you reference the issues if we have any?

nkzawa:
I just found this bug, but it seems https://github.com/zeit/styled-jsx/issues/95 is the same.

cgood92:
I pulled this down and tried in in my project - it's fixed a few problems I have had.  

Thanks @nkzawa!  I hope that this get's merged soon.... please! :)

cgood92:
@rauchg: Do we need to do a version bump on this so that I can get this using npm/yarn?  Or publish?  Right now when you pull `npm install` 0.5.1, these changes are not in there...

PR #99, opened 2017-01-28
Fix pseudo selectors and classes

giuseppeg:
This branch fixes pseudo selector and classes being suffixed when they should not.

```
:root {}
::before {}
:hover {}
p :hover {}
```

should not be suffixed (except for the p ofc)


giuseppeg:
🤔 actually they should be suffixed probably then we may want to whitelist only a few of them like `:root` – it is not quite clear to me how to use such a selector, probably making it global is even wrong.

giuseppeg:
I will discard this PR and investigate / think further

PR #100, opened 2017-02-01
[wip] Transpile external css

### LINKED ISSUE #83: Allow for importing styles
Sometimes styles can be quite long and it makes sense for them to live in a seperate file. For example, if you are using [next.js](https://github.com/zeit/next.js) and you want to put your styles in a `./styles.js` file next to a component (minimal styling just for the example):
```
export default `
  p {
    color: red;
  }
`
```
Then import it:
```

export default () => (
  <div>
    <p>only this paragraph will get the style :)</p>
    <style jsx>{styles}</style>
  </div>
)
```

Will give you the following error: `Module build failed: SyntaxError: Expected a template literal or String literal as the child of the JSX Style tag (eg: <style jsx>{`some css`}</style>), but got Identifier`.

giuseppeg:
fixes #83

### DIFF
--- a/test/fixtures/external-styles.js
+++ b/test/fixtures/external-styles.js
@@ -1,0 +1,7 @@
+const foo = 'red'
+
+export default (
+  <style jsx>{`
+    p { color: ${foo} }
+  `}</style>
+)
--- a/src/babel.js
+++ b/src/babel.js
@@ -41,7 +41,7 @@ export default function ({types: t}) {
       .get('openingElement')
       .get('attributes')
       .filter(
-        (path) => path.get('name').node.name === 'href'
+        (path) => path.get('name').node.name === 'src'
       )[0]
 
     return attr && attr.get('value').node.value
--- a/src/babel.js
+++ b/src/babel.js
@@ -21,6 +21,21 @@ export default function ({types: t}) {
     name && name.name === GLOBAL_ATTRIBUTE
   ))
 
+  const isExternalFile = ({parentPath}) => {
+    if (t.isExportDefaultDeclaration(parentPath.node)) {
+      return true
+    }
+
+    let path = parentPath.parentPath
+    if (t.isExpressionStatement(path.node)) {
+      path = path.get('expression')
+    }
+    return (
+      t.isAssignmentExpression(path.node) &&
+      path.get('left').getSource() == 'module.exports'
+    )
+  }
+
   const isStyledJsx = ({node: el}) => (
     t.isJSXElement(el) &&
     el.openingElement.name.name === 'style' &&
@@ -32,8 +47,15 @@ export default function ({types: t}) {
   const findStyles = path => {
     if (isStyledJsx(path)) {
       const {node} = path
-      return isGlobalEl(node.openingElement) ?
-        [path] : []
+      if (isGlobalEl(node.openingElement)) {
+        return [path]
+      } else if (isExternalFile(path)) {
+        const ret = [path]
+        ret.__isExternal = true
+        return ret
+      } else {
+        return []
+      }
     }
 
     return path.get('children').filter(isStyledJsx)
@@ -125,7 +147,7 @@ export default function ({types: t}) {
     }
   }
 
-  const makeStyledJsxTag = (id, transformedCss, isTemplateLiteral) => {
+  const getInjectableCss = (transformedCss, isTemplateLiteral) => {
     let css
     if (isTemplateLiteral) {
       // build the expression from transformedCss
@@ -143,7 +165,11 @@ export default function ({types: t}) {
       css = t.stringLiteral(transformedCss)
     }
 
-    return t.JSXElement(
+    return css
+  }
+
+  const makeStyledJsxTag = (id, css) => (
+    t.JSXElement(
       t.JSXOpeningElement(
         t.JSXIdentifier(STYLE_COMPONENT),
         [
@@ -161,7 +187,7 @@ export default function ({types: t}) {
       null,
       []
     )
-  }
+  )
 
   return {
     inherits: jsx,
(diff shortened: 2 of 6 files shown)


giuseppeg:
@rauchg @nkzawa early feedbacks are appreciated, what do you think about [the syntax](https://github.com/zeit/styled-jsx/pull/100/files#diff-8a986913c0f530c2109e5c1463876158R4)? should we stick with the `style` tag instead?

rauchg:
1. Isn't `src` better?
2. How would we keep track of the file for hot code reloading?

giuseppeg:
> Isn't src better?

I am going for `<style jsx src="./styles.js" />`

> How would we keep track of the file for hot code reloading?

We don't have to it should happen automatically because the external *stylesheet* is a regular js module. Not sure if I am missing something though but I will able to tell soon – I am half way through implementing the import injection and style tag replacement :)

giuseppeg:
I thought a little bit about this approach and I think that it could work in case we are cool with allowing only a single style tag when using external css:

```jsx
/* valid */

export default () => (
  <p>
      test
      <style jsx src="./style.js" />
  </p>
)

/* not valid */

export default () => (
  <p>
      test
      <style jsx>{`p { font-size: 3em }`}</style>
      <style jsx src="./style.js" />
  </p>
)
```

This is because I would use the resolved `src` path for hashing.

@nkzawa @rauchg if you like this solution and think that this feature could add value to the library I will finalize it and put in a mergeable state.

nkzawa:
I think it can work even if there are multiple style elements since we already support with `global`  ?

test/fixtures/external-styles.js
```diff
@@ -0,0 +1,7 @@
+const foo = 'red'
+
+export default (
+  <style jsx>{`
+    p { color: ${foo} }
+  `}</style>
```
nkzawa: Does this syntax mean we can't import others from imported styles ? 🤔

giuseppeg:
> I think it can work even if there are multiple style elements since we already support with global ?

when external files are transpiled they don't have knowledge of the consumer component therefore when transforming the css we don't have all of the necessary information to create a proper hash (jsxId).

```jsx
/* external file */

`p { color: red }`
```

The `jsxId` here is ```hash(`p { color: red }`) === 'woot'```
i.e. `p[data-jsx="woot"] { color: red }`

```jsx
/* consumer */

`p { font-size: 3em }` // own styles
src="./style.js"
```

The `jsxId` in this case will be ```hash(externalIdReference + `p { font-size: 3em }`) === 'wat'```

i.e.  `p[data-jsx="wat"] { font-size: 3em }`
and elements are scoped using `wat`: `<p data-jsx="wat">test</p> `
external styles won't apply because they have id `woot` instead.

> Does this syntax mean we can't import others from imported styles ?

Uh this is kinda tricky, I am using that syntax so that I can tell if a module is an external styled-jsx sheet. If we don't do so then we'd need a separate plugin to transpile external files.

giuseppeg:
The only way I can think to allow multiple style elements is to transpile external css like so:

```css
p[data-jsx~="woot"] { color: red }
``` 

(notice the `~`)

and the consumer eventually will be:

```css
p[data-jsx~="wat"] { font-size: 3em }
```

and

```jsx
<p data-jsx="woot wat">test</p>
```

rauchg:
The other thing I've been thinking about is that we could have "flow" analysis. This would work:

```jsx
import style from './style'
export default (
  <div>
    <p>hi</p>
    <style jsx>{style}</style>
  </div>
)
```

and this would work:

```jsx
cons style = `
  p { color: red }
`
export default (
  <div>
    <p>hi</p>
    <style jsx>{style}</style>
  </div>
)
```

giuseppeg:
🤔 I thought about case 1 for external files but in order to do so we'd need to hash using the resolved file name. Also when transpiling an external file we need a way to tell if that's just an external stylesheet or any other module/react component. That's why in my wip I export JSX (a style tag).

rauchg:
Why can't we just hash based on contents? The idea is that by analyzing the "flow" we can access the raw string _exactly the same_ as if it was inlined inside `<style jsx>`. Hence, IMHO, why the approach is elegant

rauchg:
It also seems like a _natural complement_ of the work we've been doing to allow constants inside the source

tgoldenberg:
@rauchg based on .jsx syntax, my first reaction was to automatically *assume* that you could do this `<style jsx>{style}</style>`. The fact that it does not do this goes against basic React syntax conventions. Also, not being able to load multiple style tags seems counter-intuitive, as well. If we are going to break people's expectations, we should make it very clear in the beginning, in order to avoid unnecessary confusion. Will be really excited to see the solution in action, and great work!

rauchg:
@tgoldenberg all of those problems are tractable. We tend to ship and iterate.

test/fixtures/external-styles.js
arunoda: Why don't we simply for a template string like:

```
const foo = 'red'

export default `
  p { color: ${foo} }
`
```

giuseppeg:
Don't get me wrong I actually like the idea of flow analysis.

> The idea is that by analyzing the "flow" we can access the raw string exactly the same as if it was inlined inside <style jsx>.

This means that we'd need to read external files as we transpile and in general the traversal would get a bit trickier. Also we may need to revisit the expressions logic since at the moment we just replace them with placeholders before hashing (we don't analyse them).

In the end though probably this will result in the better api.

test/fixtures/external-styles.js
giuseppeg: This would be the best api however is not that easy to achieve this because in the consumer file (the one that imports this) we need to hash ``` `p { color: ${foo} }` ``` and once we have the hash we need to transform the css (add the prefix to it etc).

giuseppeg:
as I said maybe for external files we can hash the resolved path instead of the content (css)

PR #101, opened 2017-02-01
Inline flex fix

cgood92:
Anything with the right hand value that contained `flex` would be replaced with `flex` ONLY.  So, `inline-flex` was being replaced to `flex`.  

With this little regex solution, you can now have both `inline-flex` and `flex` work, it will just use what it finds there.

Tested this locally, and it seemed to fix the issue.


lib/style-transform.js
```diff
@@ -241,9 +241,10 @@
                     }
                     // display: d, i, s
                     else if (first === 100 && second === 105 && third === 115) {
-                        if (line.indexOf('flex') > -1) {
```
giuseppeg: I think that we can avoid the regexp by doing the following:

```js
const flexIndex = line.indexOf('flex')
if (flexIndex > -1) {
  const flexValue = line.charAt(flexIndex-1) === '-' ? 'inline-flex' : 'flex'
  line = 'display:-webkit-'+flexValue+'; display:'+flexValue+';'
}
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "styled-jsx",
-  "version": "0.5.2",
+  "version": "0.5.3",
```
giuseppeg: we usually do this on master, can you revert?

giuseppeg:
@cgood92 thank you!

PR #102, opened 2017-02-02
Don't debounce rendering on Safari

nkzawa:
This PR fixes that FOUC occurs on Safari and iOS Safari.

Tested with Chrome, Safari, Firefox, iOS Safari, Android Chrome

Related: https://github.com/zeit/styled-jsx/issues/52

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -5,6 +5,11 @@ const update = typeof window === 'undefined' ? doRender : updateOnClient
 let components = []
 let updatePromise
 
+// includes desktop and iOS, based on
+// https://github.com/DamonOehlman/detect-browser/blob/master/lib/detectBrowser.js
+const isSafari = typeof window === 'undefined' ?
+  false : /Version\/[0-9._]+.*Safari/.test(window.navigator.userAgent)
+
 export default class extends Component {
   componentWillMount() {
     mount(this)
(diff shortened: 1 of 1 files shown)


giuseppeg:
@nkzawa did you try to use `setTimeout` in Safari?

src/style.js
```diff
@@ -5,6 +5,11 @@ const update = typeof window === 'undefined' ? doRender : updateOnClient
 let components = []
 let updatePromise
 
+// includes desktop and iOS, based on
+// https://github.com/DamonOehlman/detect-browser/blob/master/lib/detectBrowser.js
+const isSafari = typeof window === 'undefined' ?
+  false : /Version\/[0-9._]+.*Safari/.test(window.navigator.userAgent)
```
giuseppeg: I have no idea how robust is this regexp but ideally you could use [useragent](http://npmjs.com/package/useragent) which is quite battle tested or check their regexps if the package affects our lib's size.

nkzawa:
Both `setTimeout` and `requestAnimationFrame` don't work. I think we don't want to use `useragent` just for detecting Safari, and Regexp looks similar anyway.

https://github.com/3rd-Eden/useragent/blob/master/lib/regexps.js#L1188

PR #108, opened 2017-02-03
Fix jsx id bug when using interpolation in template string.

### LINKED ISSUE #107: Different components get the same jsx id if using interpolation.
For example:

``` javascript
// header.js
const Header = () => (
  <div className="root">
    <style jsx>{`
      .root {
        position: fixed;
        height: ${theme.height};
      }
    `}</style>
  </div>
);

// main.js
const Main = () => (
  <div className="root">
    <style jsx>{`
      .root {
        padding-top: ${theme.height};
      }
    `}</style>
  </div>
);
```

Both of them using interpolation in styles, and they get the same jsx id `[data-jsx="3533501187"]`.

lixiaoyan:
Fix #107

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -275,7 +275,7 @@ export default function ({types: t}) {
             ])
           }
 
-          state.jsxId = hash(state.styles.map(s => s[1]).join(''))
+          state.jsxId = hash(state.styles.map(s => s[1].source || s[1]).join(''))
           state.hasJSXStyle = true
           state.file.hasJSXStyle = true
           // next visit will be: JSXOpeningElement
--- a/test/fixtures/different-jsx-ids.js
+++ b/test/fixtures/different-jsx-ids.js
@@ -1,0 +1,23 @@
+const color = 'red'
+const otherColor = 'green'
+
+const A = () => (
+  <div>
+    <p>test</p>
+    <style jsx>{`p { color: ${color} }`}</style>
+  </div>
+)
+
+const B = () => (
+  <div>
+    <p>test</p>
+    <style jsx>{`p { color: ${otherColor} }`}</style>
+  </div>
+)
+
+export default () => (
+  <div>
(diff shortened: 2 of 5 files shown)


src/babel.js
```diff
@@ -275,7 +275,7 @@ export default function ({types: t}) {
             ])
           }
 
-          state.jsxId = hash(state.styles.map(s => s[1]).join(''))
+          state.jsxId = hash(state.styles.map(s => s[1].source || s[1]).join(''))
```
giuseppeg: maybe this could just be `hash(state.styles.reduce((h, s) => h+=s[0], ''))`

giuseppeg: i.e. use `styleId` instead of `styleText` to generate the hash

giuseppeg: that might blow up all the test and require us to refactor them, in that case maybe we can stick with your solution

giuseppeg:
@lixiaoyan good catch and thanks for the PR, can you add a test for this use case?

lixiaoyan:
@giuseppeg I think `test/fixtures/expressions.js` is enough, I have no idea to write a test for and only for this bug.

giuseppeg:
> I have no idea to write a test for and only for this bug

@lixiaoyan something like your example from #108 should do it. They should render with different jsx id

giuseppeg:
@lixiaoyan thank you!

giuseppeg:
@nkzawa can you release 0.5.4 ?

cgood92:
Thank you @lixiaoyan.... this has been messing me up for hours today, and I was just about to submit this same PR.  I pulled down the latest, saw that you had already done what I was just about to do, kicked myself for not checking, and then wanted to thank you 👍

rauchg:
We've all been there @cgood92. Thank you for thinking about helping out `styled-jsx`!

PR #110, opened 2017-02-06
Accurately document what it transcompiles to

sheerun:
By reading the source, it seems you use `styleId` instead of `data-jsx` on _JSXStyle


giuseppeg:
@sheerun good catch! thank you

Issue #115, opened 2017-02-08
Expressions don't work with prefixed properties


fatfisz:
Long story short: the output generated from this code

```js
const animationDuration = '200ms';

<style jsx>{`p { animation-duration: ${animationDuration} }`}</style>
```

contains `___styledjsxexpression_0___`. Only the first occurence of `___styledjsxexpression_0___` is substituted with the proper expression.

PR #116, opened 2017-02-08
Fix expressions for the autoprefixed properties (#115)

fatfisz:
This fixes the issue described in #115.

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -333,7 +333,7 @@ export default function ({types: t}) {
             transformedCss = css.replacements.reduce(
               (transformedCss, currentReplacement) => {
                 transformedCss = transformedCss.replace(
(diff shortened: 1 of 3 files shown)


giuseppeg:
@fatfisz looks good. Do you have multiple references to `___styledjsxexpression_0___` because of autoprefixing?

fatfisz:
Yup, that's the conclusion I came to. Or is there a possibility of a property being multiplied even if not because of autoprefixing?

fatfisz:
I mean, this only happened for `animation`, `animation-duration`, and not for `width` or even `unknown-property`.

giuseppeg:
Right now I cannot thing of any other case. Thanks for the PR

Issue #8, opened 2016-12-05
Update Compiler Fork


thysultan:
There's a few things missing pre-v0.6.4 of the stylis compiler that your fork will not handle that could be nice to have.

1. flat css
```
/* this should get wrapped in the namespace */
color: red;

h1 {
    font-size:1px;
}
```

2. full prefix support in `@root and @keyframes blocks`. 

3. support opting out of name-spacing keyframes and animations.

I've done away with the regex solution i was using, and added supports for `stylis("[data-id=namespace]")` namespaces.

rauchg:
@thysultan what do you think about `:host` support instead?

thysultan:
@rauchg  So these will be the same`& { ... } === :host { ... }` ?

rauchg:
yep so

```
<div>
  <style jsx>{`
    :host {
      color: red
    }
  `}</style>
</div>
```

`:host` would impact the div.

Alternatively, we could make the `div` also targetable simply with `div { }`, but I'm not sure if that's elegant

rauchg:
cc @giuseppeg

thysultan:
sure i can add that, `:host` will be an alias for `&`.

giuseppeg:
@rauchg correct `:host` should resolve to the *shadow root* aka the `div` in the example. 

One thing to keep in mind is that styles applied from the outside should win over `:host`'s but I guess that this behavior is hard to shim.

Nice to have then it would be to get `:host-context(selector)` support e.g.:

```jsx
<body>
  <div>
    <style jsx>{`
      :host-context(body) {
        color: red
      }
    `}</style>
  </div>
</body>
```

yields

```css
body div[data-jsx="abc"] { color: red }
```

But probably it is better to tackle this later and maybe also consider [the `._aB1bC-ext` thing](https://github.com/zeit/next.js/issues/249#issuecomment-262178503) that I suggested in the original issue.

thysultan:
@rauchg @giuseppeg pushed a patch for all the host variations `:host, :host(body), :host-context(body)`[commit](https://github.com/thysultan/stylis.js/commit/6efa027998daa107f14b0218b7c5b4440693dbe7#diff-5d9c96c5d55156f1be8b4a4d6a934a58R195)

giuseppeg:
@thysultan cool in order for `* {}` to work you need to append the namespace though otherwise `${namespace}* {}` won't work.

thysultan:
[...]

PR #125, opened 2017-02-13
[wip] Stylis CSS Middleware initial commit

giuseppeg:


lib/style-transform.js
```diff
@@ -1,407 +1,126 @@
...
+  // Stylis adds the namespace at the beginning
+  if (str.indexOf(prefix + ' ') !== 0) {
+    return
+  }
+
+  // while want to append it to selectors.
+  // Strip stylis' namespace.
+  str = str.substring(prefix.length).trim();
+
+  var i = 0
```
thysultan: @giuseppeg Everything after this point should already be handled by stylis and the [middleware](https://github.com/zeit/styled-jsx/issues/8#issuecomment-278991605), Is there a reason to iterate through the characters again?

giuseppeg: @thysultan in your snippet you split selectors by space `str.split(' ')` which is not reliable.

thysultan: @giuseppeg Apart from `[title="hello world"] h2` an attribute case what other case do you think that will break on. I have the following that will make it work in the above case.

```js
if (pos !== -1) {
	var edge;

	if (/\[.*[ ].*\]/g.exec(str) !== null) {
		edge = true;
		str = str.replace(/ (.*(?=\]))/g, '<-!->$1');
	}

	var parts = str.split(' ');

	for (var i = 0; i < parts.length; i++) {
		var part = parts[i];
		pos = part.indexOf(':');

		parts[i] = pos !== -1 ? part.substring(0, pos) + prefix + part.substring(pos) : part + prefix;
	}

	str = parts.join(' ');

	if (edge === true) {
		str = str.replace(/<-!->/g, ' ');
	}

	return str;
}
```

thysultan: This might be better than the split solution i first came up with.

```js
if (ctx === 1.5) {
	// stylis does not prefix `:global()`
	if (str.indexOf(prefix+' ') === 0) {
		// remove prefix
		str = str.substring(prefix.length).trim();

		// single selector
		if (str.indexOf(' ') !== -1) {
			return str.replace(/.+\]|[^ \n\r\t]+/g, function (match) {
				var indexOf = match.indexOf(':');

				// :hover etc...
				if (indexOf !== -1) {
					return match.substring(0, indexOf) + prefix + match.substring(indexOf);
				}
				else {
					return match + prefix;
				}
			});
		}
		else {
			return str + prefix;
		}
	}
}
```

giuseppeg: at this stage my version seems more robust (I get less compilation errors)

thysultan: Ok, will test it out locally to see what tests fail, might be vendor prefixing and whitespace issues since stylis has gone through a few patches since.

thysultan: @giuseppeg Tweaked the middleware a bit only thing failing now is source-maps generation [https://github.com/thysultan/styled-jsx/commits/master](https://github.com/thysultan/styled-jsx/commits/master) last two commits.

giuseppeg: cool! do normal and `:global` selector support nested parenthesis? e.g. `:global(div:not(.foo))`

giuseppeg: if you create a branch and submit a PR (even if work in progress) I can leave some comments on your code

thysultan: > do normal and :global selector support nested parenthesis? e.g. :global(div:not(.foo))

Yes.

```
h1 :global(div:not(.foo)) {
	color: red;
}
```

produces...

```
h1[data-jsx="123"] div:not(.foo) {color: red;}
```

giuseppeg:
we went for #134

PR #127, opened 2017-02-15
[WIP] rewrite css compiler

thysultan:
cc @giuseppeg


lib/style-transform.js
```diff
@@ -1,407 +1,45 @@
...
-                line += styles[i];
-            }
-
-            // next character
-            i++;
-            columnNumber++;
+const stylis = require('stylis')
+
+function middleware(ctx, str, line, col, prefix) {
+if (ctx === 1.5) {
```
giuseppeg: can you avoid nested blocks and return early?

```js
if (ctx !== 1.5) {
  return
}
```

and so on... (same below)

lib/style-transform.js
```diff
@@ -1,407 +1,45 @@
...
+  // stylis does not prefix `:global()`
+  if (str.indexOf(prefix) === 0 && str.indexOf(' ') !== -1) {
+    // remove prefix
+    str = str.substring(prefix.length).trim();
+
+    var indexOf;
+
+    // single selector
+    if (str.indexOf(' ') !== -1) {
+      return str.replace(/^[^ \t]*?\[.*?\]|[^ \n\r\t\+\~\>]+|.+\]/g, function (match) {
```
giuseppeg: can you move this regexp to a variable and add a comment to explain what it is doing?
Ideally it would be nice to break it down and explain every part of it.

Also you can use `\s` to match space characters (new lines and tab included I think) and `\S` for non-space chars

lib/style-transform.js
```diff
@@ -1,407 +1,45 @@
...
+        // :hover etc...
+        if ((indexOf = match.indexOf(':')) !== -1) {
+          // :global()
+          if (match.charCodeAt(indexOf+1) === 103) {
```
giuseppeg: can you use `charAt` instead? it is easier to follow if we use the actual character instead of its code

thysultan: i tend to lean towards`charCodeAt` since it's faster to compare numbers, i'll add the necessary comments so it's clear what character it's looking for.

lib/style-transform.js
thysultan: sure, and the function and regex can be hoisted.

lib/style-transform.js
thysultan: Since the idea is to also add source-maps support will use a switch

```js
switch (ctx) {
    1.5: {
        // selector middleware
    }
    1:
    2: {
        // source-map generation
    }
}
```

thysultan:
@giuseppeg added the changes though`.test[title=","]` was failing, patched it in stylis, so it's ok now. TODO: adding source-maps support.

lib/style-transform.js
giuseppeg: maybe ifs are still better so we avoid "nesting pyramids" :)

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
+  }
+}
+
+function middleware (ctx, str, line, col) {
+  if (ctx !== 1.5) {
+    return
+  }
+
+  // stylis does not prefix `:global()`
+  if (str.indexOf(prefix) === 0 && str.indexOf(' ') !== -1) {
```
giuseppeg: early return

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
-            columnNumber++;
-        }
-
-        return output;
+function replace (match) {
+  // :hover etc...
+  if ((indexOf = match.indexOf(':')) !== -1) {
+    // tries to matche `g` character in :global()
+    if (match.charCodeAt(indexOf + 1) === 103) {
+      return match.replace(regExpGlobal, '$1').trim()
```
giuseppeg: I have mixed feelings about regexp globals, maybe you can use a function as 2nd argument for `replace`?

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
+
+  // stylis does not prefix `:global()`
+  if (str.indexOf(prefix) === 0 && str.indexOf(' ') !== -1) {
+    // remove prefix
+    str = str.substring(prefix.length).trim()
+
+    // single selector
+    if (str.indexOf(' ') !== -1) {
+      return str.replace(regExpReplace, replace);
+    } else {
```
giuseppeg: you don't need an `else` if you are returning above

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
+      if ((indexOf = str.indexOf(':')) !== -1) {
+        return str.substring(0, indexOf) + prefix + str.substring(indexOf);
+      } else {
```
giuseppeg: same, you don't need the `else`

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
+  }
+}
+
+module.exports = function (id, styles) {
+  prefix = `[data-jsx="${id}"]`
 
-    return stylis;
-}));
+  return stylis(prefix, styles, true, false, middleware)
+}
```
giuseppeg: add an empty line at the end of the file

test/fixtures/transform.css
```diff
@@ -121,3 +129,11 @@ test
...
   display: flex;
 }
+
+.test {
+  box-shadow: 0 0 10px black, inset 0 0 5px black
+}
+
+.test[title=","] {
+  display: inline-block;
+}
```
giuseppeg: please add an empty line at the end of the file

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
-
-            // next character
-            i++;
-            columnNumber++;
-        }
-
-        return output;
+function replace (match) {
+  // :hover etc...
+  if ((indexOf = match.indexOf(':')) !== -1) {
```
giuseppeg: you can invert the logic here and have

```js
if ((indexOf = match.indexOf(':')) === -1) {
  return match + prefix
}
```

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
+    // tries to matche `g` character in :global()
+    if (match.charCodeAt(indexOf + 1) === 103) {
+      return match.replace(regExpGlobal, '$1').trim()
+    } else {
```
giuseppeg: don't need an `else` here

lib/style-transform.js
```diff
@@ -1,407 +1,77 @@
...
-                line += styles[i];
-            }
-
-            // next character
-            i++;
-            columnNumber++;
-        }
-
-        return output;
+function replace (match) {
```
giuseppeg: what is this function replacing? can you give it a more self explanatory name?

thysultan:
@giuseppeg submitted updates and add source-map... At this point i don't know why package.json is conflicting, i get no conflicts when i try to rebase locally, what am i missing?

giuseppeg:
@thysultan are you pulling/rebasing against this repo or origin (your fork)?

Try

```bash
git remote add upstream https://github.com/zeit/styled-jsx.git
git pull upstream master
```

thysultan:
@giuseppeg Thanks, forgot to pull upstream, conflicts resolved now. On this branch `generates source maps` passes on my end, though the test was failing on master before i started this branch. Thoughts?

rauchg:
LGTM

leo:
@thysultan Very good PR, thanks! 😊

However, I had to remove the commit from master, because we forgot to rebase your branch before merging it. This overwrote some commits we made before.

Could you please rework the commits on master? If not, I'll do it for you and commit.

PR #129, opened 2017-02-16
Allow whitespace in expressions inside template literals

### LINKED ISSUE #126: Allow whitespace in expressions inside template literals
As of now expressions like the following won't get replaced because I'm using spaces  inside the curly braces.

```JavaScript
<script jsx>{`
  p {
    color: ${ colors.primary }
  }
`}</script>
```
It has to be written as `${colors.primary}`.

Also note that the README's [demo](https://github.com/zeit/styled-jsx#constants-and-config) is using that style.

I believe this is the corresponding source: https://github.com/zeit/styled-jsx/blob/45ba8542f93d5caee262ead08a584a25767eacd5/src/babel.js#L108.

nikvm:
Fixes #126.

So far I refrained from adding an extra test case and instead extended the already present `expressions` fixture.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "babel-traverse": "^6.21.0",
     "babylon": "^6.14.1",
     "convert-source-map": "^1.3.0",
+    "escape-string-regexp": "^1.0.5",
     "object.entries": "^1.0.4",
     "source-map": "^0.5.6",
     "string-hash": "^1.1.1"
--- a/src/babel.js
+++ b/src/babel.js
@@ -5,6 +5,7 @@ import {SourceMapGenerator} from 'source-map'
 import convert from 'convert-source-map'
 import traverse from 'babel-traverse'
 import {parse} from 'babylon'
+import escapeStringRegExp from 'escape-string-regexp'
 
 // Ours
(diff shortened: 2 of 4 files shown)


giuseppeg:
@nikvm thank you! Do you think that we should add an extra test for strings that contain characters that need to be escaped?

nikvm:
Do you mean: _need to be escaped for the `RegExp`_?

If so, we already have a fixture testing [`p { ${darken(color) + 2} }`](https://github.com/zeit/styled-jsx/blob/master/test/fixtures/expressions.js#L14) where `(`, `)` and `+` need to be escaped. If that was still leading to bugs with escaping it would mean [`string-escape-regexp`](https://www.npmjs.com/package/escape-string-regexp) (which we use internally for escaping) doesn't properly do it's job and tests/fixes should be implemented over there. :trollface:

I guess returning the question: What would a test like that look like?

giuseppeg:
> If that was still leading to bugs with escaping it would mean string-escape-regexp (which we use internally for escaping) doesn't properly do it's job and tests/fixes should be implemented over there.

fair enough :)

giuseppeg:
thank you man!

PR #131, opened 2017-02-16
Add README section on Syntax Highlighting

### LINKED ISSUE #114: Syntax highlighting with atom & language-babel
Enable css syntax highlighting in atom when using language-babel package: gandm/language-babel#324

nikvm:
I thought this might close #114. Info taken from [here](https://github.com/gandm/language-babel/issues/324).

I'll be doing some research into how this could be accomplished in other editors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,24 @@ It's **paramount** that you use one of these two functions so that
 the generated styles can be diffed when the client loads and
 duplicate styles are avoided.
 
+## Syntax Highlighting
+
+When working with template literals a common drawback is missing syntax highlighting. The following editors currently have support for highlighting CSS inside `<style jsx>` elements.
+
+ _If you have a solution for an editor not on the list_ __please [open a PR](https://github.com/zeit/styled-jsx/pull/new/master)__ _and let us now._
+
+### Atom
+
+The [`language-babel`](https://github.com/gandm/language-babel) package for the [Atom editor](https://atom.io/) has an option to [extend the grammar for JavaScript tagged template literals](https://github.com/gandm/language-babel#javascript-tagged-template-literal-grammar-extensions).
+
+After [installing the package](https://github.com/gandm/language-babel#installation) add the code below to the appropriate settings entry. In a few moments you should be blessed with proper CSS syntax highlighting. ([source](https://github.com/gandm/language-babel/issues/324))
+
+```
+"(?<=<style jsx>{)|(?<=<style jsx global>{)":source.css.styled
+```
+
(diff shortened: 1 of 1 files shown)


giuseppeg:
Awesome! I was thinking of adding a FAQ section at the bottom of README.md with a few `<summary>` blocks to answer the most common / related questions, like Next.js does https://github.com/zeit/next.js#faq .

To give you an idea every once somebody asks about this

```
Warning: Unknown prop `jsx` on <style> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in style (created by Unknown)
    in nav (created by Unknown)
    in Unknown (created by Unknown)
    in div (created by Unknown)
    in Unknown (created by App)
    in App
```

which means that for some reason styled-jsx didn't run completely (eg. they are not using next@2.0.0-beta)

nikvm:
Being blunt, I'm very much against FAQs. :gun:

It's a comfortable place to stuff in content that seemingly doesn't fit in anywhere else. In my opinion this only pushes the problem to the end user (who honestly doesn't know which questions are _frequently asked_). I'd rather think a little longer about where said content belongs and add appropriate sections if necessary. Note that __this is very much subjective__. But it's hard to get rid of a FAQ section once it's there [[1]](#user-content-next-faq).

The proposed _Syntax Highlighting_ section might seem superfluous at the moment due to it having only one entry. But I'm hoping it will grow over time as we find solutions for other editors. I also believe it's important enough to deserve it's own section.

As for your example: It might fit in a section _Common Pitfalls_ or better yet in the [Next.js](https://github.com/zeit/next.j) docs. So we can link to it when this type of error comes up.

In any case: No :heart: feelings. In the end it's not that big of a deal.

<a href="#user-content-next-faq" name="next-faq">[1]</a>: I'm working on [documentation improvements for Next.js](https://github.com/zeit/next.js/issues/1069). And I'm trying to get rid of the FAQ section for above reasons. Also I don't want to step on anybody's toes and would be happy to back away.

nikvm:
To conclude: I wouldn't wait for more editors to make it to the list before merging. Personally I'm using [VS Code](https://code.visualstudio.com/) where it's currently not supported. If you'd rather  add it as a FAQ, feel free&mdash;or I can take care of that. :v:

I'd just like to see this documented somewhere and most&mdash;importantly&mdash;have the satisfaction of closing an issue. :metal:

giuseppeg:
That's totally ok and makes sense. There are actual FAQ like the example I mentioned, but it is not a big deal we can add a paragraph for that later.

PR #132, opened 2017-02-16
Minify Global Styles

### LINKED ISSUE #120: Global CSS is not minified
`<style jsx>`
```
<style id="__jsx-style-3974295953">.link[data-jsx="2236456407"] {background: red;}</style>
```

`<style jsx global>`
```
                <style id="__jsx-style-1864081020">
                    .link {
                        background: blue;
                    }
                </style>
```

nikvm:
An attempt to fix #120.

My first thought was to implement a minifier (or import one). But we already got our transformer!
So why not send the global styles through without assigning an id (and adding a suffix to selectors)?
On further thinking about it, it even seems more _correct_. As of now global styles don't get transformed at all which means no Browser prefixing etc.

Am I missing something? :hear_no_evil:

Obviously tests don't pass, yet. Also proposed changes need a bit of refactoring/ commenting if accepted. :shit:

Update: Sorry, a bit to many commits...


src/babel.js
```diff
@@ -297,7 +297,8 @@ export default function ({types: t}) {
           const [id, css, loc] = state.styles.shift()
 
           if (isGlobal) {
-            path.replaceWith(makeStyledJsxTag(id, css.source || css, css.modified))
+            const transformedCss = transform(null, (css.source || css))
```
giuseppeg: nothing wrong with this but you could also move `let transfromedCss` (L306) to L298

giuseppeg:
@nikvm we rewrote the css compiler so when you have time you should rebase your changes.

nikvm:
Still on it, by the way. Sorry for the delay 🌴

nikvm:
This should to the trick.

test/fixtures/mixed-global-scoped.out.js
```diff
@@ -1,8 +1,8 @@
 import _JSXStyle from 'styled-jsx/style';
-const Test = () => <_JSXStyle styleId={188072295} css={"p { color: red }"} />;
+const Test = () => <_JSXStyle styleId={188072295} css={" p {color: red }"} />;
 
 export default (() => <div data-jsx={793889750}>
     <p data-jsx={793889750}>test</p>
-    <_JSXStyle styleId={3149549172} css={"body { background: red }"} />
+    <_JSXStyle styleId={3149549172} css={" body {background: red }"} />
```
giuseppeg: we should suppress those leading whitespaces

src/babel.js
```diff
@@ -295,16 +295,15 @@ export default function ({types: t}) {
...
           let transformedCss
+          const useSourceMaps = Boolean(state.file.opts.sourceMaps)
 
-          if (useSourceMaps) {
+          if (isGlobal) {
+            transformedCss = transform(
+              null,
+              (css.modified || css)
+            )
+          } else if (useSourceMaps) {
```
giuseppeg: We should have source maps for global styles too I think.

Maybe you can do something like this:

```js
if (useSourceMaps) {
   transformedCss = [
       transform(
          isGlobal ? null : String(state.jsxId),
          // ...
       ),
       // ...
   ]
} else {
      transformedCss = transform(
         isGlobal ? null : String(state.jsxId),
         // ...
      )
}
```

test/fixtures/mixed-global-scoped.out.js
nikvm: Right, I noticed that, too. Will fix.

src/babel.js
nikvm: Didn't even think about that. Will fix.

nikvm: ☄️ 0f7b202

test/fixtures/mixed-global-scoped.out.js
nikvm: I might just `trim()` the transformed css.

giuseppeg: it should be fixable here https://github.com/zeit/styled-jsx/blob/master/lib/style-transform.js#L120-L128

either try to do `blob = blob.trim()` before L120 or move L127 to L119

giuseppeg: I think that the space is there because of this https://github.com/thysultan/stylis.js/blob/master/stylis.js#L905

thysultan: @nikvm actually you might want to just add to [L120](https://github.com/zeit/styled-jsx/blob/master/lib/style-transform.js#L120)

```js
if (prefix.length === 0) {
    return blob.trim()
}
```

since it's using an 0 length string for global allowing the middleware to only handle non-global selectors.

nikvm: Thanks for the tip @thysultan. I committed the fix. 🐛

giuseppeg:
@nikvm awesome, the PR looks good thank you! Did you have the chance to test source maps?

nikvm:
@giuseppeg I added a global style tag to the source maps fixtures [here](https://github.com/nikvm/styled-jsx/blob/14f1b3a0c9adf12e2f06e60c99dd90cf2e2ed93d/test/fixtures/source-maps.js#L6) and [here](https://github.com/nikvm/styled-jsx/blob/14f1b3a0c9adf12e2f06e60c99dd90cf2e2ed93d/test/fixtures/source-maps.out.js#L6). But didn't check the generated mapping string.

rauchg:
Would love to merge this! Just need to update the tests. Thanks for all your help so far @nikvm

fatfisz:
Hi, this would be great to have in styled-jsx, but it's been not merged for two months. Is there something I could do to help get it going?

giuseppeg:
@fatfisz after #148 this branch need to be reworked. I'll try to include a fix in the next release.

giuseppeg:
fixed in #194 #198

PR #134, opened 2017-02-18
rewrite css compiler

thysultan:
cc @leo


giuseppeg:
@thysultan 👍 you've got to fix `package.json`. Your branch is bringing back unpinned dependencies. My suggestion is that you copy/paste it from master and add `stylis` (pin the version to `1.1.11`. Once you've done that remove `node_modules` and `dist` and reinstall. Tests should fail. For some reason it seems that the current version of the source maps generator is not adding `charset=utf-8;` anymore – I guess it is fine to remove it?

thysultan:
@giuseppeg fixed, and removed `charset=utf-8;`

giuseppeg:
awesome @thysultan thank you!

PR #137, opened 2017-02-26
isSafari should include webviews

jgautheron:
I was pulling my hair for a while with this issue, until I noticed that this bug was recently [fixed](https://github.com/zeit/styled-jsx/pull/102) but the browser sniffing was not including iOS webviews.


giuseppeg:
Thank you @jgautheron! we had to remove batch styles updates so I will close this PR.

jgautheron:
Even better 👍 
@giuseppeg A release for this fix would be great!

PR #139, opened 2017-02-27
Remove batched styles updates

### LINKED ISSUE #136: Batched styles updates don't immediately affect DOM nodes
This issue was initially opened in zeit/next.js/issues/1239 

Quoting myself: 

> @nkzawa this is an issue with rAF. I believe that the issue is with styles updates being async and therefore the computed element width is wrong because it's not been affected by styles yet.

The OP is trying to access the `clientWidth` of a DOM node in `componentDidMount`. 

A simplified test case would be:

```js
const body = document.body
requestAnimationFrame(() => {
  body.style.backgroundColor = 'red'
})
console.log(getComputedStyle(body).backgroundColor)
```

Result

```
rgba(0, 0, 0, 0)
```

I am not sure if we can fix it without removing batch updates. 
Ideas are welcome @rauchg @nkzawa @ericf

giuseppeg:
Fixes #136

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -1,14 +1,7 @@
 import {Component} from 'react'
 import render from './render'
 
-const update = typeof window === 'undefined' ? doRender : updateOnClient
 let components = []
-let updatePromise
-
-// includes desktop and iOS, based on
-// https://github.com/DamonOehlman/detect-browser/blob/master/lib/detectBrowser.js
-const isSafari = typeof window === 'undefined' ?
-  false : /Version\/[0-9._]+.*Safari/.test(window.navigator.userAgent)
 
 export default class extends Component {
   componentWillMount() {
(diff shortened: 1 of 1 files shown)


giuseppeg:
@nkzawa @rauchg ping

nkzawa:
👍

though not so sure if you really need to get values from styled components on React.

iamjacks:
@nkzawa I noticed the bug when I was making a responsive slider. I'd imagine if anyone is working on something more robust with responsive design and/or animations they may run into this issue as well. #140 is related. In my case I didn't mind moving my components style to the global space to remedy the problem. Everyone's mileage varies though.

chrillo:
@iamjacks I would argue that even if it might do harm its still rather important as it "breaks" the promises one comes to rely on when working with the component live cycle methods mainly the component being there and ready when componentDidMount gets called. based on you comment does adding the global flag <style jsx global> fix the issue? if so I could also live with that solution

giuseppeg:
By adding the global flag you lose scoped styles which is not acceptable imho. 

Also style tags are blocking for a reason and in this case the user should get the expected behavior. I think that you could work around this issue by deferring your logic with `setTimeout` but again that's not the expected behavior.

@rauchg @nkzawa if you don't have any objections I will merge this PR

chrillo:
@giuseppeg thanks for the clarification. yes setTimeout is exactly what I used to work around this issue.

nkzawa:
@giuseppeg I'm 👍 . I think we can reconsider when have any performance problems.

PR #143, opened 2017-03-02
Add Webstorm/Idea support tip.

phtoussaint:
This only provides temporary support, but it's very easy to action and incredibly useful.
Edit: I'm looking for a way to set it as a rule but could not find a good documentation for IntelliLang which is quite complicated. I'm happy with the current solution.


phtoussaint:
See demo.

styled-jsx-webstorm-inject-language

rauchg:
This is awesome. Thanks for your help

Issue #106, opened 2017-02-03
processor for stylelint


lixiaoyan:

giuseppeg:
I believe that stylelint can be added if you use [styled-jsx-postcss](https://github.com/giuseppeg/styled-jsx-postcss) instead.

@rauchg I suppose that this is never going to happen in `styled-jsx` right? stylelint is based on PostCSS.

thysultan:
@giuseppeg what about [https://csstree.github.io/docs/validator.html](https://csstree.github.io/docs/validator.html)?... or when the transformer is updated to v1.0 stylis you could register a custom linter as a plugin.

rauchg:
It'd be cool to have support for linting, with stylis

thysultan:
@rauchg working on v1.2 that will have better support for creating a linter plugin, will create a basic linter plugin after that is out.

thysultan:
@rauchg This should now be possible with v1.2.0 [relevant issue](https://github.com/thysultan/stylis.js/issues/8)

giuseppeg:
I think that this can be done as a plugin if/when #190 lands. Can we close it?

rsternagel:
@lixiaoyan

Got it working!

On top of stylelint you need the "{at}mapbox/stylelint-processor-arbitrary-tags" processor and the "stylelint-config-css-modules" configuration (for allowing "global()"-syntax).

```
yarn add stylelint --dev
yarn add stylelint-config-standard --dev 
yarn add @mapbox/stylelint-processor-arbitrary-tags --dev
yarn add stylelint-config-css-modules --dev
```
XOR

```
npm i stylelint --save-dev
npm i stylelint-config-standard --save-dev
npm i @mapbox/stylelint-processor-arbitrary-tags --save-dev
npm i stylelint-config-css-modules --save-dev
```


### .stylelintrc
```javascript
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-css-modules"
  ],
  "processors": [
    [ "@mapbox/stylelint-processor-arbitrary-tags", {
            "startTag": "\\s*<style jsx>{`",
            "endTag": "\\s*`}<\/style>"
      }
    ]
  ],
  "rules": {
    "no-empty-source": null
  }
}
```

### Now lint code:
```
./node_modules/.bin/stylelint "src/**/*.js"
```

Of cource, this can be done nicer with a npm script...

IMO it's worth adding to the "readme.md". :)

dmitrika:
[...]

Issue #103, opened 2017-02-03
Make semicolon optional?


bassbump:
Since with ES6 and all of zeits packages it's become standard to not use the semicolon, wouldn't it be consistent to also apply this practise in the styled-jsx?

giuseppeg:
We are not using semicolons in js. What do you mean?

bassbump:
```
<style jsx>{`
  .page-header {
    background-color: #000
    height: 60px
  }
`}</style>
```

instead of 
```
<style jsx>{`
  .page-header {
    background-color: #000;
    height: 60px;
  }
`}</style>
```

nkzawa:
Semicolon of css is not optional except last declarations in blocks.

giuseppeg:
ditto

Issue #93, opened 2017-01-24
unexpected behaviour of :global()


davibe:
I would expect `.root :global(p + p) { ...` to work but it does not.

```
const Bug = () =>
  <div className="root">
    <p>red</p>
    <p>green</p>
    <p>blue</p>
  <style jsx>{`

  .root :global(p) {
    color: red;
  }

  /* this does not work */
  .root :global(p + p) {
    color: green;
  }

  /* this works */
  .root :global(p) + :global(p) {
    color: blue;
  }

  `}</style>
  </div>
```

To make things more clear i am creating a Typography element with css rules that are scoped to the `.root` element but are applied to all sub-element of sub-component. I do this to style markdown-rendered texts.

giuseppeg:
`div :global(> *)` doesn't work too – it gets compiled to 

```
div[data-jsx="3533501187"] > *)[data-jsx="3533501187"]
```

giuseppeg:
should be fixed by #134

michalwerner:
I have `styled-jsx` from master branch with #134 but I still can't do anything more than single selector in prefixed `:global`. For example:

this works
```
<style jsx>{`
  .menu :global(li) {
    color: red;
  }
`}
```

this does not work:
```
<style jsx>{`
  .menu :global(ul li) {
   color: red;
  }
  .menu :global(ul > li) {
   color: red;
  }
  .menu :global(li + li) {
    color: red;
  }
`}
```

but this also works (no `.menu` prefix):
```
<style jsx>{`
  :global(ul li) {
    color: red;
  }
`}
```

Is this intended behavior?

thysultan:
This is a bug, fixed upstream(stylis v1.2.2).

rauchg:
@thysultan should we bump?

thysultan:
@rauchg I'll send a PR.

PR #144, opened 2017-03-05
fix :global & animations

thysultan:
- fixes [unexpected behaviour of :global()](https://github.com/zeit/styled-jsx/issues/93) and another issue that surfaces when animation* is the last property without a semicolon i.e 

```scss
h1 {animation: foo}
// should generate
h1 {animation: foo}
// was generating 
h1 {animation: foo;
```

- bumped stylis to version 1.2.3 this should make [linter plugins](https://github.com/zeit/styled-jsx/issues/106) and [https://github.com/zeit/styled-jsx/issues/103](https://github.com/zeit/styled-jsx/issues/103) possible.


rauchg:
Thank you so much for all your awesome work @thysultan

Issue #135, opened 2017-02-19
Add prettier


giuseppeg:
~~Currently we use [xo](https://github.com/sindresorhus/xo) to lint our code.
We should replace it with [standard.js](https://github.com/feross/standard)~~

Let's just add Prettier. See https://github.com/zeit/now-cli/blob/master/package.json#L39-L52

nkzawa:
I wonder if we can just format by `https://github.com/jlongster/prettier` instead. @rauchg

giuseppeg:
We can have "prettier" fixtures but we still need a linter for the library code

giuseppeg:
oh wait I thought prettier was only for JSX, awesome!

rauchg:
prettier + linting rules for mistakes / warnings

nikvm:
I started experimenting with this.

My starting point was to lint using [`eslint-config-standard`](https://github.com/feross/eslint-config-standard) as base and resetting styling rules with [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier). The idea was then to auto-fix styling on a [pre-commit hook](https://github.com/prettier/prettier#pre-commit-hook-for-changed-files) and only lint for common mistakes. This way a dev isn't bothered at all with errors like `there should be a space at line 22 column 68`. Also vanilla `eslint` makes it easy to use editor plugins for real time feedback.

__The problem__: I noticed `prettier` adds semicolons to the output which might not be configurable any time soon (discussion [here](https://github.com/prettier/prettier/issues/736)). There is a workaround for this, like piping `prettier`'s output into `standard --fix` (There is a [package](https://github.com/dtinth/prettier-standard-formatter) for that). In this case code gets formatted once using `prettier`'s advanced algorithm and then again using `standard --fix`. Which&mdash;due to different styling rules&mdash;does more than just removing semicolons. In my opinion this gets more complicated than it needs to be.

This might not be a big problem. But before I continue I wanted to check what the general consent is or rather what the expected developer experience is concerning linting and (auto-fixing) code style.

My suggestion is to use `prettier` as is or just use `standard` by itself (which again leaves a dev with styling errors, also fixable by  using `--fix` somewhere in the chain).

📓 You may experiment with the changes [here](https://github.com/nikvm/styled-jsx/tree/feature/pretty-standard). After cloning run `yarn install` followed by `npm run lint` and `npm run prettier` or `npm run prettier-standard`. Diff the output.

cc @giuseppeg @rauchg

[...]

PR #145, opened 2017-03-05
[W.I.P.] Replace `xo` with `standard.js` and `prettier`

nikvm:
Fixes: #135 

* Removes `xo`
* Adds `eslint` with a subset of `standard`, styling rules removed
* Adds `prettier --write` and `eslint` pre-commit hook for staged files
* Allows customization of `prettier` styling via [`prettier-eslint`](https://github.com/prettier/prettier-eslint), for now:
  * No semicolons
  * Single quotes

__Notes:__

* The pre-commit hook rewrites files in place. [There is a warning in the `prettier` docs for doing this](https://github.com/prettier/prettier#cli)
* Linting will catch "bad code" plus the styling rules we configure (i.e. semicolons, quotes)
* I'm not yet sure if it's the best idea to add `standard`, then remove parts of it and add some of them again. Haven't found another solution yet, unless we're ok with `prettier` styling as it is.
* It's configured to lint everything but test fixtures and `lib/` code. Should we add/remove anything?


rauchg:
I think I'd prefer `eslint:recommended`, so that no style rules are imposed? Or maybe there's a way of putting `standard` in such a mode? Thanks

nikvm:
Actually I thought about the same. At the moment I use [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) which removes `standard` style rules (i.e. anything that conflicts with `prettier`). That said, `standard ` does more than `eslint:recommended` regarding code quality like adding `import` and `promise` plugins and configuring [`parserOptions`](http://eslint.org/docs/user-guide/configuring#specifying-parser-options) which we would have to maintain our self. Also this might not be a bad thing. I guess there is always some tradeoff.

rauchg:
Yeah that's an excellent point. I've been finding myself adding a lot of extra rules to `eslint:recommended`.

orlin:
@nikvm @rauchg I was trying to do the same today with `prettier-eslint` + `standard`, so I figured out a couple of things that may help.

If you reverse the order of eslint config `extends` to `["prettier", "standard"]` the standard rules (coming last) don't get removed, so you won't need the explicit `quotes` and `semi` nor any other overrides (e.g. space before function parentheses).  Seems like everything works as before.

I also found the [eslint-plugin-prettier](eslint-plugin-prettier) useful.  Example eslint config:

```
"plugins": [
    "prettier"
  ],
"rules": {
    "prettier/prettier": [0, {
      "singleQuote": true,
      "trailingComma": "all",
      "jsxBracketSameLine": true
    }]
  }
```

This fixed missing semicolon errors that [prettier-atom](https://github.com/jlongster/prettier-atom) gave (guessing a bug).

orlin:
I guess the second point about `eslint-plugin-prettier` could be irrelevant, as `prettier-eslint` has a `logLevel` option that would probably do the same...

nikvm:
@orlin reversing like you suggested means we're, again, linting for _styling rules_ and has the same effect as not extending `eslint-config-prettier` at all.

```
config-standard + config-prettier = standard rules - styling rules
config-prettier + config-standard = standard rules

```

`eslint-config-prettier` is only there to __override__ `standard` styling rules.

Also concerning @giuseppeg's comment [here](https://github.com/zeit/styled-jsx/issues/135#issuecomment-284336281) I feel like I might not have made myself clear what the desired outcome of this change is.

### Here is where I see the problem and what I like to accomplish

Traditionally when linting code we do so for two reasons: Consistent __code style__ (e.g. spaces vs. tabs) and __code quality__ (e.g. no unused variables).

With the introduction of `prettier` we don't have to be concerned with _code style_ at all anymore and&mdash;in my opinion&mdash; shouldn't lint for it. Instead enforce a rewrite through pre-commit hooks so that any versioned code is guaranteed to have the same style without manual intervention or even needing to test for it. 😍 

This leaves us with _code quality_ which we absolutely should be linting for. There is two routes I propose. First, use a well established configuration like `standard` minus _code style_ rules which pushes maintenance to the maintainers of said package.  Or alternatively, configure _code quality_ rules ourself (e.g. use `eslint:recommended`) which gives us more freedom but also more maintenance overhead and again more reason for discussion about what rules do include. 🗡 

Of course, in order to call our self _Real JS Dev&trade;_, we have to discuss semicolons. Unfortunately `prettier` doesn't have an option, yet, to not print them which leaves us with the only option of adding a linting rule, if we so desire. My own aesthetics tell me to leave semicolons out of code. We just have to be aware that, for the time being, it adds some technical debt (and unnecessary linting errors). 🙉 

💫 💃

package.json
```diff
@@ -21,37 +21,38 @@
...
+    "precommit": "lint-staged",
+    "test": "npm run lint && ava",
+    "lint": "eslint  \"{.,src/**,test}/*.js\"",
+    "format": "prettier-eslint --write \"{.,src/**,test}/*.js\""
+  },
+  "lint-staged": {
+    "{.,src/**,test}/*.js": [
+      "prettier-eslint --write",
+      "eslint",
+      "git add"
```
giuseppeg: should be this done by hand?

giuseppeg:
@nikvm I think that what you've got is ok. If you want to keep this simple just replace xo with standard and we can reconsider to add prettier when they add support for no-semicolons

orlin:
@nikvm I basically agree with what you’re saying, though I’ll probably keep the `extends: [“prettier", "standard”]` because I just don’t want to invest the time to research what the rules difference is.  My guess is that even with the semicolon option (if / when that happens) there would still be other diffs that someone would have to reconcile and maintain a config patch with something like `prettier-standard` or `standard-prettier` — I wonder which that would be :)

Meanwhile, my bottom line is that prettier does improve the formatting even with the standard config changing some of its style rules.  Using `prettier-atom` I have Format on Save enabled, so the format on commit hook is only there for contributors not using an editor with `prettier-eslint` integration.

Once in a while after upgrades, I may flip the order of prettier & standard configs and run `yarn format` to see what has changed.

sheerun:
@orlin [`prettier-standard`](https://github.com/sheerun/prettier-standard) it is. it also supports custom eslint configuration

nikvm:
@giuseppeg If ok with core contributors we can use it as is. I enjoy the auto-fixing quite a lot. Keeps the mind on the more important aspects of writing code.

package.json
sheerun: no, it just re-adds already staged file after fixing its formatting

giuseppeg:
@nikvm just saw this https://twitter.com/notquiteleo/status/844134256418930688 :) Maybe we can do the same? cc @leo

leo:
@giuseppeg I'd love that. But internally, we came to the conclusion that the author of a repository will be the one deciding which linting tool to use. And since I didn't build styled-jsx from the ground up (@rauchg and others did), he'll be the one deciding.

Generally, I think my solution is:

- The best one, because XO is much stricter than "standard"
- The cleanest one, because we're only inheriting the "prettier" config, so custom rules

nikvm:
ℹ️ I've been away without a proper Internet connection and will be until June. I won't be able to work on this PR until then. If someone wants to take over, please feel free.

❤️ :octocat:

giuseppeg:
fixed in #196

PR #148, opened 2017-03-12
Add external StyleSheets support

### LINKED ISSUE #83: Allow for importing styles

giuseppeg:
Fixes #83

This branch introduces external stylesheets support.

External stylesheets are js modules that export a string or template literal:

```js
const color = 'red'

export default `
  p { color: red } 
`
```

Which can be used as follow:

```jsx
import styles from './styles'

export default () => (
  <p>
     Zeit rocks
     <style jsx>{styles}</style>
  </p>
)
```

Please read the 

<details>
<summary>
Implementation details
</summary>
Say that you have `components/Card/style.js`:

```
export default `
 div { color: red }
`
```

File names are unique so we can generate the hash for external files using the filename `hash(path.resolve('components/Card/style.js'))` and transpile the file like this:

```
export default `
 div[data-jsx=ext~="woot"] { color: red }
`
```

the `~` matches a word therefore you can have `data-jsx-ext~="woot wat"` where `woot` and `wat` are external stylesheets hashes.

```
<div data-jsx="wut" data-jsx-ext="woot wat">
 ...
 <_JSXStyle id="wat" css="div[data-jsx=\"wat\"] {font-size: 2em}" />
 <_JSXStyle id=”woot" css={referenceToTheImportedFile} />
</div>
```

For the sake of supporting both global and local styles the transpiled stylesheet exports an object with two keys `global` and `local` respectively for the original version (global) and scoped one. This allows people to use the same stylesheet in both modes (maybe it is overkill and we can remove it if you want).
</details>

### TODOs

- [x] Transpile external stylesheets that pass the CSSTree validation
- [x] Add support for source maps
- [x] Add support for ES5/module.exports
- [x] Test edge cases and maybe in a real app
- [x] Add support for named exports
- [x] Use external files' content for hashing and export the hashes instead of hashing the filename 
- [x] Add DOCs to README.md


giuseppeg:
You guys may want to review per commit since the branch is quite big :)
Also do you know how to suppress the linter error (it is the only reason why tests fail)?

fatfisz:
@giuseppeg For that linter error you can just use elision:
```js
const [
  id,
  ,
  externalStylesReference,
  isGlobal
] = state.externalStyles.shift()
```

giuseppeg:
Test are passing locally and failing in CI because `resolve('./foo.js')` resolves to different absolute paths in CI, locally. Can you guys think of a workaround?

rauchg:
wow. Amazing progress

giuseppeg:
@rauchg since babel processes files alphabetically (I believe) external stylesheets are not transpiled in some cases. An example would be `button.js` imports some styles defined in `a.js`.

To work around this issue we can take one of the following approaches:

1. We ask the user to annotate the file e.g. by adding a comment at the top of the file `/* @styled-jsx */`
2. We only support external stylesheets whose extension is `jsx` this way we can treat the file as a stylesheet when it exports a string.
3. We ask the users to use a sub extension like `.styled.js`
4. We try to detect CSS modules by doing some checks on the exported string – in this case I'd write a module `is-css-string` maybe. This is the best solution in terms of DX but it can be error prone.

src/babel-external.js
```diff
@@ -0,0 +1,124 @@
...
+    return
+  }
+  const css = getExpressionText(path)
+  const prefix = `[${MARKUP_ATTRIBUTE_EXTERNAL}~="${styleId}"]`
+  const isTemplateLiteral = Boolean(css.modified)
+  const useSourceMaps = Boolean(state.file.opts.sourceMaps)
+
+  let globalCss = css.modified || css
+
+  if (validate && !isValidCss(globalCss)) {
```
giuseppeg: if we want to avoid this we could require people to tag external stylesheets like so:

```js
export default styledJsx`
  div { color: red }
`
```

rauchg:
> since babel processes files alphabetically (I believe) external stylesheets are not transpiled in some cases

I wonder if there's a way to invoke babel from within the plugin to work around this limitation? As long as caching is enabled, it would incur in overhead for the user (double parsing)

giuseppeg:
@rauchg I am not sure if 1) plugins get the dest path 2) if we write the file to disk after transforming it with babel-core (like we do for tests) that would conflict or be overridden by the original one (when transpiled)

giuseppeg:
@rauchg @nkzawa this branch is ready for review, I still need to write some docs and I'll test this branch with next-news.

Here is a breakdown of what I did:

### usage 

External stylesheets are ES6 or CommonJS modules that export a string or template literal.

```jsx
/* styles.js */
export default `div { background-color: red }`
```

```jsx
/* otherStyles.js */
module.exports = "p { color: white }"
```

Expressions work.

Within components these can be referenced like so:

```jsx
import styles from './styles'
const otherStyles = require('./otherStyles')

export default () => (
  <div>
     <p>hello</p>
     <style jsx>{styles}</style>
     <style jsx>{otherStyles}</style>
  </div>
)
```

Styles can also be consumed as `global` by adding the attribute to the `style` element as usual.

### how this works

Instead of hashing the CSS string, I calculate the hash for each external stylesheet by using its absolute path (filename) which is always unique. 

```js
hash(filaname)
```

Since one may want to use multiple stylesheets I introduced a new data attribute `data-jsx-ext` and I am using the following attribute selector in css:

```css
[data-jsx-ext~="a"] { /* ... */ }
[data-jsx-ext~="b"] { /* ... */ }
```

Notice the `~=`. Such a selector matches any element whose `data-jsx-ext` attribute value is a list of space-separated values, one of which is exactly equal to `a` or `b`.

Meaning that in the markup we will have:

```jsx
<div data-jsx-ext="a b c d">
```

That said the external stylesheets are rewrote to something like this

```js
export default {
  global: `div { background-color: red }`,
  local: `div[data-jsx-ext~="597123001"] { background-color: red }`
}
```

And the `style` tags rewritten like so:

```jsx
<_JSXStyle styleId={597123001} css={styles.local} />
<_JSXStyle styleId={597123003} css={styles.global} />
```

### implementation details

You may want to review things by branch – maybe it'd make it easier to follow what is going on.

* I moved all of the utils and constants to separate files
* Added logic to detect external stylesheets references and transpilation – here is where I resolve the import source (filename), hash it and add it to the elements
* Transform `style` tags and transpile external stylesheets (and add the `{ global, local }` thing). 

Note at this point I was keeping track of the external stylesheets references and transpile those files as I encountered them. However this has proven to be wrong because: depending on the order in which babel processes files some stylesheets might never been compiled. This is due to the fact that `styles.js` occurs (and is processed) before `ZooComponent.js` for example.

* I solved this issue by validating exported strings and template literals with CSSTree: if a module exports a string or a template literal then I pass that string through CSSTree and if parsing doesn't fail then it must be CSS. [CSSTree is very fast](https://github.com/postcss/benchmark#parsers).

An alternative solution would be to tag the file with a comment `/* @styled-jsx */` or wrap the exported strings/literal in a function (identity) ```export default styledJsx(`div { color: red }`)```.

* Finally since `hash(filename)` is unique on each machine I had to strip hashes out from tests. We could find a more cool way to mock/stub the hash but I guess that his is fine.

Alternatively probably we can hash the whole file content instead of the filename and remove this issue. 

If you need any more info you know where to find me :)

giuseppeg:
> Alternatively probably we can hash the whole file content instead of the filename and remove this issue.

This would add some overhead since we'd need to do i/o (read the file from disk)

thomaslindstrom:
Just a question: Would you, with this, be able to use `webpack` loaders to import a CSS file that you inject into the `style` tag?

```javascript
import stylesheet from './styles.css'

export default () => (
  <div>
     <p>hello</p>
     <style jsx>{stylesheet}</style>
  </div>
)
```

```css
// styles.css
p {
  color: red;
}
```

giuseppeg:
@thomaslindstrom I am not exactly sure but in theory yes. 
It'd be amazing if you could clone this branch and try locally.

nkzawa:
sorry for late replay. I didn't read details of implementations yet, but ideas seem brilliant 👍

giuseppeg:
@nkzawa no worries, I just pinged you because I would like to merge this (if we think that it is a good solution) to avoid crazy merge conflicts (I touched quite a lot)

rauchg:
@giuseppeg

- do named exports work?
- do multiple imported expressions work inside the same `<style jsx>` tag?

```
import { a, b } from './external-styles'
export default () => (
  <div>
    <style jsx>{ a } { b }</style>
  </div>
)
```

giuseppeg:
@rauchg 

> do named exports work?

No they don't, only `default` exports and `module.exports` right now. 
This is to avoid parsing every export and validating every string or template literal with csstree. If you think that it is a useful feature I can add support for it.

> do multiple imported expressions work inside the same <style jsx> tag?

Not right now because of [this check](https://github.com/zeit/styled-jsx/blob/master/src/babel.js#L242-L246), we could add support for that too though. 
This would also enable mixed local and external styles support.

```jsx
import { a } from './external-styles'
export default () => (
  <div>
    <style jsx>
       {`div { color: red }`}
       { a }
    </style>
  </div>
)
```

tz5514:
Looks great! Hope that release soon.

Larry850806:
I'm looking foward to see this PR be merged.

lededje:
What are we waiting on to get this merged? Happy to write documentation if that's all that is missing.

giuseppeg:
Hold in the excitement folks! :D We'll merge (soon) once this branch has been reviewed and is good to go :)

lib/style-transform.js
```diff
@@ -34,128 +34,118 @@ const regExpReplaceGlobal = /:global\((.*)\)/g
...
-  }
-
-  // post-processed
-  if (context === 6 && generator !== void 0) {
-    generator = void 0,
-    offset = void 0,
-    filename = void 0
-
-    return
+function transform (prefix, styles, settings = {}) {
```
giuseppeg: fyi this is the only important change I made to the stylis middleware.

giuseppeg:
@rauchg, I had a chat with @nkzawa and the outcome was the following:

- [ ] as you suggested we should add support for named exports
- [ ] we are going to create hashes from the content and export the hash instead of hashing the filename. The latter is problematic since the full file path is generally unique so CI or any machine other than ours would generate different hashes.
```jsx
import styles from './styles'
// ...
<_JSXStyle id={styles.hash} css={styles.local} />
```

Finally we need your opinion on the public and private api for external stylesheets.
Internally right now I transpile strings to an object with two keys `global` and `local`.

```js
export default `div { color: red }`
// becomes
export default {
  global: `div { color: red }`,
  local: `div[data-jsx-ext~="123"] { color: red }`
}
```

This is so that people can consume those styles both as scoped (local) and global. 

@nkzawa doesn't like this api that much and I must admit that I am not a big fan too because now the original module exports are lost. This could be a potential issue if other modules are relying on them.

Should we ignore the double mode thing (global and local) and just return the scoped styles as string?

On the public api side instead @nkzawa asked whether we can do

```jsx
export default <style jsx>{`div { color: red }`}</style>
```

This currently works for global styles because we don't need to rewrite jsx tags – see [example](https://github.com/zeit/styled-jsx/blob/master/test/fixtures/mixed-global-scoped.js#L1). 

My answer to that was, yes it is doable and it would simplify the implementation but probably it is not great for DX (developers experience).

```js
export default `div { color: red }`
``` 

is much better imho. 

The cons of this syntax though is that we need to do CSS detection i.e. if a module (any module) exports a string we should parse that string with csstree (super fast) to determine whether it is a valid css string.

I am for making the DX a priority at the cost of having to do CSS detection.

giuseppeg:
@rauchg @nkzawa ok made the changes above, please look at the snapshots (fixtures .out.js files) and let me know if they look ok. If so I'll add some docs and merge.

PR #152, opened 2017-03-23
Fix expr namespacing inside animation shorthand

soulchainer:
Changes the placeholder for expressions to avoid namespacing non
`animation-name` strings when using them inside `animation` shorthand.

Problem with actual code is that it replaces temporarily every found expression with
`__styledjsxexpression_${id}__`. This is OK in almost every CSS property... but nope in
`animation` shorthand. Because, if you use and expressión to set a value different than
the `animation-name`, stylis.js will namespace it, thinking it's an `animation-name`.
Because the resulting string is a valid `animation-name` string. I just add a zero at the beginning of the placeholder because, that way, it isn't a valid `animation-name`, so stylis.js won't namespace it.
And everything else works. I added another zero at the end because, you know: symmetry.

As example, from my own code, this:

```js
 .active {
  animation: ${USER_TONE_FADE_DURATION}s forwards user-activated;
}
```

will render like this:

```js
.active[data-jsx="2540933619"] {
  ...
  animation: a25409336190.5s forwards a2540933619user-activated;
}
```

And well, the animation will be wrong-formed and won't be applied.

Of course, you can easily avoid this bug by not using the shorthand property.
But the shorthand is nice and it's there to use it when you want, right?

**Note**: it's there any prize 🥇 for a dumb PR, because I know this deserve it ^_^U.

I suspect (xD) this is hideous and ugly xD, but it's a quick & easy patch from someone not familiar with this project.
IDK if this is OK, but sure this is a problem that should be tackled, so here is my quick try :þ.

**Note 2**: there was also a little errata into the original code comment, because `___styledjsxexpression0___`, wich, according to the original placeholder, should be `___styledjsxexpression_0___`.

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -102,11 +102,11 @@ export default function ({types: t}) {
     // e.g.
     // p { color: ${myConstant}; }
     // becomes
-    // p { color: ___styledjsxexpression0___; }
+    // p { color: 0__styledjsxexpression_0__0; }
 
     const replacements = expressions.map((e, id) => ({
       pattern: new RegExp(`\\$\\{\\s*${escapeStringRegExp(e.getSource())}\\s*\\}`),
-      replacement: `___styledjsxexpression_${id}___`,
+      replacement: `0__styledjsxexpression_${id}__0`,
       initial: `$\{${e.getSource()}}`
     })).sort((a, b) => a.initial.length < b.initial.length)
 
--- a/src/babel.js
+++ b/src/babel.js
@@ -102,11 +102,11 @@ export default function ({types: t}) {
     // e.g.
     // p { color: ${myConstant}; }
     // becomes
-    // p { color: 0__styledjsxexpression_0__0; }
+    // p { color: %%styledjsxexpression_0%%; }
 
     const replacements = expressions.map((e, id) => ({
       pattern: new RegExp(`\\$\\{\\s*${escapeStringRegExp(e.getSource())}\\s*\\}`),
-      replacement: `0__styledjsxexpression_${id}__0`,
+      replacement: `%%styledjsxexpression_${id}%%`,
       initial: `$\{${e.getSource()}}`
     })).sort((a, b) => a.initial.length < b.initial.length)
 
--- a/test/fixtures/expressions.js
+++ b/test/fixtures/expressions.js
@@ -24,7 +24,7 @@ export default () => (
     }</style>
     <style jsx>{`p { animation-duration: ${animationDuration} }`}</style>
     <style jsx>{`
-      p { animation: ${animationDuration} forwards ${animationName } }`
+      p { animation: ${animationDuration} forwards ${animationName} }`
     }</style>
   </div>
 )
--- a/test/fixtures/expressions.js
+++ b/test/fixtures/expressions.js
@@ -2,6 +2,7 @@ const color = 'red'
 const otherColor = 'green'
 const mediumScreen = '680px'
 const animationDuration = '200ms'
+const animationName = 'my-cool-animation'
 
 export default () => (
   <div>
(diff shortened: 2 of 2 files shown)


giuseppeg:
@soulchainer good catch! maybe we can use `%%styledjsxexpression_0%%` instead? Can you also add a test for this case? https://github.com/zeit/styled-jsx/blob/master/test/fixtures/expressions.js

soulchainer:
@giuseppeg :D. I can add the test in a moment, but... the format you suggest doesn't work ^_^U. In fact, I thought previously in replacing the symbol delimiters. It still gets namespaced by stylis.js. It works if you add a 0 (well, any number, in fact) You can test other formats you would prefer directly using the playground stylis.js has. If you have stylis.js forked, there is a `docs` directory with an `index.html` you can open locally in your browser, with the playground. If you made an animation, as example:

```css
  animation: %%styledjsxexpression_0%%s forwards user-activated;
```

inside, you can see how stylis.js still namespaces both, the `animation-name` and the placeholder.

Another placeholder style that works, now that I did some tests, is one delimited by parenthesis, so if you liked more something like `(styledjsexpression_0)` or `(__styledjsexpression_0__)`, that works ;).

What do you think?

giuseppeg:
> I can add the test in a moment, but... the format you suggest doesn't work

The format I suggested is not a valid animation name and therefore it shouldn't prefix it i.e. stylis is buggy in that sense :)

Regarding the second issue, in general expressions won't be preprocessed unfortunately. This is a limitation of using expression that can only be fixed if we decide to preprocess them at runtime. For now though I wouldn't bother trying to find a solution in this PR.

soulchainer:
> The format I suggested is not a valid animation name and therefore it shouldn't prefix it i.e. stylis is buggy in that sense :)

Yeah, I know. I wasn't implying that it wasn't working because you did it wrong. It should work. I also tried previously because I thought the same thing. But nope. Only works with numbers >_<. It also works using parenthesis (in the playground), but then, it raises other issues with the code in styled-jsx. So... numbers, ugly delimiters xD.

And yeah, I guessed the second issue. Wasn't complaining, I know that decission has a reason. But, ideally, that would solve it (doesn't mean it should be done, things aren't usually that easy ^^).

> For now though I wouldn't bother trying to find a solution in this PR.

This mean... I should stil continue and add a test (for the case that works with this patch) or you're talking about leave it? Just to be sure, my english isn't perfect.

In any case, thanks for your patience, @giuseppeg  👍.

giuseppeg:
@soulchainer your english is great! ;)

My suggestion would be that we prepare a branch with tests where we change the placeholder syntax to `%%styledjsxexpression_N%%` (maybe add a commet to explain why it is like that) and then open an issue on the Stylis repo to fix the parser so that invalid animation names are not prefixed. 

Once that fix is merged we merge yours into styled-jsx.

thysultan:
Thanks for finding this @soulchainer, patched in stylis v1.2.7

soulchainer:
I'm fine with it, @giuseppeg :).

But still, have you thought about what I previously explained? I refer to the fact that, applying the initial fix I proposed, it would solve the bug triggering the namespacing of all expressions in `animation` shorthand (representing an `animation-name` or not), yes, but also would create a new one, preventing an `animation-name` from ever get namespaced, neither in `animation` shorthand nor in `animation-name` individual property.

It's that OK? Because I think that namespacing of `animation-name` is important, right? That's something to take into account. Needs to be tackled or solving the previous issue becomes irrelevant.

The problem is that we can't know where exactly is the `animation-name` placed in an `animation` shorthand when it's setted via an expression. Unless we force an specific order for the properties in it. Because `animation` definition is pretty flexible.

giuseppeg:
@soulchainer yes the fix I suggest is to "normalize" the behaviour. 
Right now other expressions are not preprocessed either, so you're right animation names won't be namespaced and that's _ok_. Later we have to fix this issue for all of the expressions.

giuseppeg:
@soulchainer thank you for your help man! I am fixing this in #189

PR #156, opened 2017-03-28
docs(readme) declare code blocks as jsx

pixelass:
Not all code blocks were declared as jsx therefore some examples had broken syntax highlighting


pixelass:
Example of broken syntax highlighting:

screen shot 2017-03-28 at 2 45 23 pm

giuseppeg:
awesome! thank you @pixelass!

PR #168, opened 2017-04-08
Show how to select children of global element

timneutkens:


giuseppeg:
doesn't `div :global(.react-select .some-child-of-react-select) {}` work?

timneutkens:
`div :global(.react-select .some-child-of-react-select)` compiles down to `div[12345] .react-select .some-child-of-react-select)` Notice the ending `)`. I'm not sure if this is an issue, or by intent :smile:

giuseppeg:
oh definitely a bug! Thanks I keep saying that we should rewrite the global part :)

rauchg:
This is a good workaround until we make a permanent fix. Thanks @timneutkens

Issue #88, opened 2017-01-23
Styles doesn't get appended to the head in IE10 / IE11


roopemerikukka:
I found that if I have a component that mounts another component in example by clicking some button, the styles applied for the mounted component doesn't get applied.

Here is the example case:
```jsx

import React, { Component } from 'react'
import Head from 'next/head'

class App extends Component {

  constructor () {
    super()
    this.state = {
      showDiv: false
    }
  }

  onLinkClick () {
    this.setState({
      showDiv: !this.state.showDiv
    })
  }

  render () {
    return (
      <div>
        <Head>
          <title>Next.js with styled-jsx and IE10</title>
          <script src='https://cdn.polyfill.io/v2/polyfill.min.js' />
        </Head>
        <a href='#' onClick={(e) => { this.onLinkClick() }}>Click me!</a>
        { this.state.showDiv &&
          <SubComponent />
        }
      </div>
    )
  }
}

class SubComponent extends Component {
  render () {
    return (
      <div>
        <div className='subcomponent'>
          This and that
        </div>
        <style jsx>{`
          .subcomponent {
            background-color: black;
            color: white;
          }  
        `}</style>
      </div>
    )
  }
}

export default App

```

The subcomponent doesn't get the black background or the white text colors. The component gets the `data-jsx` attribute, but the `<style></style>` tags doesn't get appended to the pages `<head>` as they are in Chrome (or other newer browsers).

This bug occurs in both IE10 and IE11.

Here is the complete example repo: https://github.com/roopemerikukka/next.js-styled-jsx-ie10-issue

roopemerikukka:
Now I noticed that this also happens when navigating to other page. Lets say I have pages like this:

```
pages/
  index.js
  item.js
```

Now if I have an component on the index page and there is a link to item page, the item page doesn't get any styles given by `<style jsx>` tags. If I navigate there directly I get the styles, propably because of SSR.

giuseppeg:
I just tried your example on Windows 10 and it works 🤔

roopemerikukka:
With which version of IE? I've tried it with IE10 on Windows 8, IE11 on Windows 7 and IE11 on Windows 10 but none of those worked (at least with browserstack). Here is the screenshots (IE11/Win10) with link clicked and without. As you see the style tag for the subcomponent is missing.

[...]

PR #187, opened 2017-05-08
Make rendering order deterministic

### LINKED ISSUE #185: Global CSS order.
Hello,

I'm loading webfonts on scaleapi.com but I see flash of invisible font.

While I add the font-face css at the top, I think it's being ignored and I see it 3rd.

screen shot 2017-05-05 at 3 58 51 am

Thanks

Rowno:
Currently the components array gets [converted to an Object map](https://github.com/zeit/styled-jsx/blob/0545f4e6e66608389edaf6bf49f0157841887186/src/style.js#L27-L29) but when it's later [looped over](https://github.com/zeit/styled-jsx/blob/0545f4e6e66608389edaf6bf49f0157841887186/src/render.js#L23) the entries are returned in a different order. This pull request changes all the Object maps to proper Map's (which always iterate over entries in insertion order).

There's no client-side tests but from what I can tell everything is working correctly.

Note: this introduces a breaking change to `require('styled-jsx/style').flush()`, however I can't see it documented anywhere so I'm guessing it's only used internally.

Fixes #185

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -1,48 +1,40 @@
-import entries from 'object.entries'
-
-const {hasOwnProperty} = Object.prototype
-const tags = {}
-let prevStyles = {}
+const tags = new Map()
+let prevStyles = new Map()
 
 export default typeof window === 'undefined' ? renderOnServer : renderOnClient
 
 function renderOnServer() {}
 
-function renderOnClient(components) {
-  const styles = {}
-  for (const c of components) {
-    styles[c.props.styleId] = c
-  }
-
+function renderOnClient(styles) {
   patch(diff(prevStyles, styles))
-
   prevStyles = styles
 }
 
 function diff(a, b) {
-  const added = entries(b).filter(([k]) => !hasOwnProperty.call(a, k))
-  const removed = entries(a).filter(([k]) => !hasOwnProperty.call(b, k))
+  const added = Array.from(b.entries()).filter(([k]) => !a.has(k))
+  const removed = Array.from(a.entries()).filter(([k]) => !b.has(k))
   return [added, removed]
 }
 
-const fromServer = {}
+const fromServer = new Map()
 
 function patch([added, removed]) {
   for (const [id, c] of added) {
     // avoid duplicates from server-rendered markup
-    if (undefined === fromServer[id]) {
-      fromServer[id] = document.getElementById(`__jsx-style-${id}`)
+    if (!fromServer.has(id)) {
+      fromServer.set(id, document.getElementById(`__jsx-style-${id}`))
     }
 
-    tags[id] = fromServer[id] || makeStyleTag(c.props.css)
+    const tag = fromServer.get(id) || makeStyleTag(c.props.css)
+    tags.set(id, tag)
   }
 
   for (const [id] of removed) {
-    const t = tags[id]
-    delete tags[id]
+    const t = tags.get(id)
+    tags.delete(id)
     t.parentNode.removeChild(t)
     // avoid checking the DOM later on
-    fromServer[id] = null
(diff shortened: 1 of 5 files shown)


giuseppeg:
Hi @Rowno thanks for the PR!
Since we dedupe styles rendering order will never be 100% deterministic I suppose.
Also if you use `Map`s we'd need to transpile them :(

Rowno:
This doesn't require transpilation, just a polyfill which you said styled-jsx shouldn't be providing: https://github.com/zeit/styled-jsx/issues/88

giuseppeg:
right fair enough.

src/style.js
```diff
@@ -21,15 +21,17 @@ export default class extends Component {
...
 
-export function flush() {
-  const ret = {}
-
-  for (const {props} of components) {
-    ret[props.styleId] = props.css
+function componentMap() {
+  const ret = new Map()
+  for (const c of components) {
+    ret.set(c.props.styleId, c)
```
giuseppeg: Can we make this backwards compatible? i.e. `ret.set(c.props.styleId, c.props.css)`?

giuseppeg: I don't know why it is exported tbh cc @nkzawa @rauchg

Rowno: Since we're returning a Map now it's not going to be compatible regardless. Also I thought it would be best if the 2 outputs from this module returned the same data format.

giuseppeg: fair point (2) :) I'll wait a day to see if either @nkzawa or @rauchg have something to add and I'll merge. Thanks again!

src/render.js
```diff
@@ -1,48 +1,40 @@
...
+let prevStyles = new Map()
 
 export default typeof window === 'undefined' ? renderOnServer : renderOnClient
 
 function renderOnServer() {}
 
-function renderOnClient(components) {
-  const styles = {}
-  for (const c of components) {
-    styles[c.props.styleId] = c
```
nkzawa: I think we can do `if (!styles.hasOwnProperty(c.props.styleId))` check to fix the issue instead.
tho I think it would be ok if we decided to require `Map`.

Rowno: I don't think that'll fix the issue, from what I can tell the Object is iterating over its properties in shortest to longest order (the hash varies in length so adding more css randomly changes the order).

nkzawa: yeah, I'm missing out `styleId` is Number.

nkzawa: Another options:

- use Object just for remove duplicated styles and return Array.
- add prefix to ids like `'id:' + c.props.styleId` and make them string.

Rowno: `styleId` is already being converted into a String because Object properties can only be Strings and Symbols.

nkzawa: Yep, object properties are ordered as number as far as I know.

giuseppeg: well now he's using a `Map`

giuseppeg: I mean maybe we can trick some engine but I don't know if they all keep object keys in order with `'id:'+styleId`.

If there is any real benefit in switching to an ordered data structure I'd say let's merge.
As I said at the beginning order will never be 100% deterministic since we dedupe styles e.g.:

```jsx
// componentA:
<Button />

// componentB: 
<Header><Button /><Header />
```

still renders:

```
button styles
header styles
```

nkzawa: yeah, again, I'm ok if we decided to require `Map`.

Rowno: The use case is for global styles like css resets which you want to initialise at the top of your component tree and render before anything else.

giuseppeg:
@Rowno thank you again for the PR and for your patience :)

Issue #81, opened 2017-01-18
dynamic styles mode


giuseppeg:
@rauchg and I discussed about possible ways to support dynamic styles i.e. allow `props` and other dynamic values which #80 doesn't.

From slack:
> @g and I are contemplating a "dynamic" mode for styled-jsx where you pay a small penalty in performance but you can write stuff like

 ```jsx
<style jsx dynamic>{`
  p {
    width: 100%;
    color: ${someProp ? 'red' : 'blue'}
  }
`}</style>
```

Basically this mean separate static css from dynamic one, render static css with the current system and inline the dynamic one i.e. each instance will get its own styles.

In order to do so we would need a CSS parser to split things up so that the example above becomes:

```javascript
[
  {
    dynamic: false,
    css: 'p { width: 100% }'
  },
  {
    dynamic: true,
    css: 'p { color: ___styled-jsx-expression_placeholder_n_1___ }'
  }
]
```

which eventually compiles down to 

```jsx
<_JSXStyle styleId={woot} css={"p[data-jsx=\"woot\"] {width: 100%;}"} />
<_JSXStyle dynamic styleId={woot2} css={`p[data-jsx=\"woot2\"] {color: ${someProp ? 'red' : 'blue'}}`} />
```

When `dynamic` is set our component renders a regular style tag instead of [returning `null`](https://github.com/zeit/styled-jsx/blob/master/src/style.js#L21).

To parse the CSS we could write a tiny parser or use something like [CSSTree](https://github.com/csstree/csstree) since it seems to be fairly fast https://csstree.github.io/docs/validator.html (we could reuse this for the css transformation as well).

giuseppeg:
@lahmatiy showed me how this could be done with CSSTree

https://gist.github.com/giuseppeg/2699094fc7f9f355e4289f6432266a83

@thysultan would it be as easy to do so with a new Stylis middleware?

thysultan:
Could add to the existing middleware

```
var selector = '', collection = '', dynamic = {};

function middlware (context, value) {
	// selector
	if (context === 1) selector = value;

	// property, and some rule to only pull actual dynamic props
	if (context === 2) {
		if (dynamic[selector] === void 0) dynamic[selector] = [];

		dynamic[selector].push(value);

		// removes dynamic prop
		return '';
	}

	// complete
	if (context === 6) {
		selector = '';
		collection = '';
                dynamic = {};

		for (var key in dynamic) collection += key + dynamic[key].join('') + '}';

[...]

Issue #5, opened 2016-12-05
SASS transformation


rauchg:
Probably outside the scope of this module, but we could document how to make this work:

```js
<script jsx type="text/sass">{`
  p {
    color: red;
    & a {
       color: #fff;
    }
  }
`}</script>
```

thysultan:
I'm looking into adding this to stylis, ping me if you have implementation ideas.

operatino:
Having a documentation about how to add own css transformation rules would be really useful.

> Note: whoever writes this transformation must be aware that we auto-prefix already.

Btw, this is also important part missing in docs. Autoprefixing should be at least configurable to define supported browsers.

drcmda:
@rauchg @operatino why not simply do it like csjs? They allow compile time babel transforms:

Links:
https://github.com/tizmagik/react-csjs
https://github.com/rtsao/babel-plugin-csjs-postcss

That way any postCSS transform can run over the css. You could safely remove the autoprefixer as its no longer needed.

thysultan:
So i already got a working version of this for stylis that i'll push with the next release and it's only like ~40 lines so i'm pretty impressed with the results, the way it works is when it finds a block it skips it but adds it as a block to the end of the string buffer namespaced to it's parent, when the parser finally comes across it again the same thing will happen for any nested blocks within it, recursively until there are no more nested blocks.

so this...

```css
h1, div {
        color: red;

	h2, &:before {
		color: red;
	}
        
       color: blue;

	header {
		font-size: 12px;
	}
}
```

becomes this...

```css
h1, 
div { color: red; color: blue; }

h1 h2,
h1:before,
div h2,
div:before { color: red; }

h1 header,
div header{ font-size: 12px; }
```

thysultan:
^^ v0.8.0 now supports this.

rstacruz:
What I'd rather see is [postcss](https://www.npmjs.com/package/postcss) support.

More specifically, [cssnext](https://www.npmjs.com/package/postcss-cssnext) is a perfect fit for the styled-jsx paradigm, since it's basically like "babel for CSS." It takes proposed CSS additions and implements them. I'd love to see a more standard approach to CSS variables.

sidorares:
+1 to cssnext, we would avoid lot's of bugs ( like #68 ) with full-featured css transpiler. Dependency size is not an issue since it's compile time step

rstacruz:
[...]

PR #190, opened 2017-05-13
Add support for plugins

### LINKED ISSUE #182: Make styled-jsx extensible
I thought of a way to make styled-jsx a bit more flexible and let users manipulate the CSS both at compile and run time (if they want). This for example would make it possible to use SASS with styled-jsx or I could stop supporting styled-jsx-postcss.

The babel plugin could accept an option `transformCss` that is a module name. Such a module exports an object with two functions for runtime and compile time. Both signatures would be `(css: string) => string` (by default an `identity` function).

```js
// my-styled-jsx-transform/index.js
export default {
  compile: (css) => sass.renderSync({ data: css }),
  runtime: (css) => css
}
```

and in `.babelrc`:

```json
{
   "plugins": [
      ["styled-jsx/babel", { "transformCss": "my-styled-jsx-transform" }]
   ]
}
```

Which we will use as follow:

```js
const {compile: _userCompile} = require(state.opts.transformCss)
// set compile to identity function if it is `undefined`

// TBD

const transformedCss = stylis(
   styleJsxId,
   _userCompile(css)
)

// or

const transformedCss = _userCompile(
  stylis(
     styleJsxId,
     css
  )
)
```

I am not sure about the runtime one (which for now could be postponed) but probably we would need to unshift an import like we do with `_JSXStyle`. The result would be:

```jsx
import {runtime as _styledJsxTransformCss} from 'my-styled-jsx-transform'
// ...
<_JSXStyle id={1235435} css={_styledJsxTransformCss(`div { color: red }`)} />
```

cc @rauchg @nkzawa

giuseppeg:
Fixes #182

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -48,6 +48,7 @@ export default () => (
 - Runtime size of just **2kb** (gzipped, from 6kb)
 - Complete isolation: Selectors, animations, keyframes
 - Built-in CSS vendor prefixing
+- CSS Preprocessing via Plugins
 - Very fast, minimal and efficient transpilation (see below)
 - High-performance runtime-CSS-injection when not server-rendering
 - Future-proof: Equivalent to server-renderable "Shadow CSS"
@@ -112,6 +113,39 @@ module.exports = `div { color: green; }`
 // the following won't work
 // module.exports = { styles: `div { color: green; }` }
 ```
+
+### CSS Preprocessing via Plugins
+
+Styles can be preprocessed via plugins.
+
+Plugins are regular JavaScript modules that export a simple function with the following signature:
+
+```js
+(css: string) => string
+```
+
+Basically they accept a CSS string in input, optionally modify it and finally return it.
+
+Plugins make it possible to use popular preprocessors like SASS, Less, Stylus, PostCSS or apply custom transformations to the styles.
+
+To register a plugin add an option `plugins` for `styled-jsx/babel` to your `.babelrc`. `plugins` must be an array of module names or *full* paths for local plugins.
+
+```json
+{
+  "plugins": [
+    [
+      "styled-jsx/babel",
+      { "plugins": ["my-styled-jsx-plugin-package", "/full/path/to/local/plugin"] }
+    ]
+  ]
+}
+```
+
+In order to resolve local plugins paths you can use NodeJS' [require.resolve](https://nodejs.org/api/globals.html#globals_require_resolve).
+
+Plugins are applied in definition order left to right before styles are scoped.
+
+N.B. when applying the plugins styled-jsx replaces template literals expressions with placeholders e.g. `%%styledjsxexpression_ExprNumber%%` because otherwise CSS parsers would get invalid CSS.
 
 ### Targeting The Root
 
--- a/readme.md
+++ b/readme.md
@@ -42,6 +42,11 @@ export default () => (
 )
 ```
 
+### Options
+
+* `sourceMaps` generates source maps (default: `false`)
+* `vendorPrefix` turn on/off automatic vendor prefixing (default: `true`)
+
 ## Features
 
 - Full CSS support, no tradeoffs in power
@@ -172,14 +177,16 @@ Each plugin receives a `settings` object as second argument which contains
 the babel and user options:
 
 ```js
-export default (css, settings) => { /* ... */ }
+(css, settings) => { /* ... */ }
 ```
 
 The `settings` object has the following shape:
 
 ```js
 {
+  // babel options
   sourceMaps: true,
+  vendorPrefix: true,
 
   // user options
   options: {
--- a/readme.md
+++ b/readme.md
@@ -121,7 +121,7 @@ Styles can be preprocessed via plugins.
 Plugins are regular JavaScript modules that export a simple function with the following signature:
 
 ```js
-(css: string) => string
+(css: string, settings: Object) => string
 ```
 
 Basically they accept a CSS string in input, optionally modify it and finally return it.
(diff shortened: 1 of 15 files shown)


readme.md
```diff
@@ -112,6 +113,39 @@ module.exports = `div { color: green; }`
...
+  "plugins": [
+    [
+      "styled-jsx/babel",
+      { "plugins": ["my-styled-jsx-plugin-package", "/full/path/to/local/plugin"] }
+    ]
+  ]
+}
+```
+
+In order to resolve local plugins paths you can use NodeJS' [require.resolve](https://nodejs.org/api/globals.html#globals_require_resolve).
```
giuseppeg: It would be nice if we could find a way to magically resolve paths internally. 
If it is not possible then asking them to use `require.resolve` or anyway provide full paths is ok I guess.

readme.md
```diff
@@ -42,12 +42,18 @@ export default () => (
 )
 ```
 
+### Options
+
+* `sourceMaps` generates source maps (default: `false`)
+* `vendorPrefix` turn on/off automatic vendor prefixing (default: `true`)
```
giuseppeg: should this be a noun/plural? `vendorPrefixes`

readme.md
```diff
@@ -113,6 +119,82 @@ module.exports = `div { color: green; }`
...
+```
+
+Each plugin receives a `settings` object as second argument which contains
+the babel and user options:
+
+```js
+(css, settings) => { /* ... */ }
+```
+
+The `settings` object has the following shape:
```
giuseppeg: this object could as well be renamed to `options` and change shape:

```js
{
  // user options
  exampleOption: true,

  // babel options
  babel: { // or env
    sourceMaps: true,
    vendorPrefix: true,
  }
}
```

readme.md
```diff
@@ -113,6 +119,82 @@ module.exports = `div { color: green; }`
...
+    ]
+  ]
+}
+```
+
+Each plugin receives a `settings` object as second argument which contains
+the babel and user options:
+
+```js
+(css, settings) => { /* ... */ }
```
nkzawa: Does it support async operations ?

nkzawa:
I wonder if we can implement the example of https://github.com/zeit/styled-jsx/issues/5. (can you get `type="text/sass"` ?)

Additionally, I wonder if there are any limitations comparing to creating a babel plugin which transforms `css` string before `styled-jsx` compilation runs.

readme.md
giuseppeg: no it doesn't because of the sync nature of babel

giuseppeg:
> I wonder if we can implement the example of #5. (can you get type="text/sass" ?)

We could but I'd rather keep those extensions in user space otherwise we have to ship styled-jsx with node-sass, postcss etc

> Additionally, I wonder if there are any limitations comparing to creating a babel plugin which transforms css string before styled-jsx compilation runs.

Definitely an option but obviously writing a simple function is easier than setting up a new babel plugin that has to detect `<style jsx>` tags, replace expressions with placeholders, preprocess the string and then replace the original template literal. Also with external styles they'd need to detect css exports like we do in order to apply plugins to external css.

nkzawa:
I'm rather wondering if you can get `props` of `<style jsx/>`. I totally agree this style of plugins would be better.

Can we have `props`  as an argument (or in `settings`) ?

```
export default (css, props, settings) => {
   if (props.type === 'text/sass') {
     return compileSass(css);
   } else {
     return css;
   }
}
```

giuseppeg:
> Can we have props as an argument (or in settings) ?

ah yes I like the idea but unfortunately we wouldn't have those info for external stylesheets :/

giuseppeg:
@nkzawa ideally people can annotate css eg:

```jsx
<style jsx>{`
  /* @styled-jsx-sass */
  div {
    color: red;
    &:hover { color: blue }
  } 
`}</style>
```

and the plugin author can check for that comment

```js
export default (css, props, settings) => {
   if (/@styled-jsx-sass/.test(css)) {
     return compileSass(css);
   } else {
     return css;
   }
}
```

readme.md
giuseppeg: fwiw in styled-jsx-postcss I use a super hack to make async code work (syncronously) https://github.com/giuseppeg/styled-jsx-postcss/blob/master/src/babel.js#L98-L107

giuseppeg:
another thing to consider is figure out if this feature could make dynamic styles (#81) harder to implement in the future

Tomekmularczyk:
what is the status of this pull request?

giuseppeg:
update: we decided to wait until we implement #81 since it might affect how plugin works.

medv:
When this gets merged I'm selling all my stuff and moving to styled-jsx island.

jabacchetta:
Also very interested in both this and #81. Trying to decide between styled-jsx or styled-components with Next.js, and these seem to be the missing pieces.

smeijer:
> @giuseppeg update: we decided to wait until we implement #81 since it might affect how plugin works.

So, what's the status of #81? Since I really like to have these plugins available, to be able to parse PostCSS.

giuseppeg:
It is in a good spot, in my little spare time I am trying to finish #81 basically there are a few minor adjustments to do, need to write extensive tests and docs. You can track the progress here https://github.com/zeit/styled-jsx/tree/dynamic-styles

Faradey27:
Hi, do we plan to merge this pull request?(and resolve conflicts)

giuseppeg:
@Faradey27 finishing up #81 and planning on updating this PR and get it merged. With the limited amount of free time that I have it might take a few weeks 😅 but I will do my best

giuseppeg:
Duplicate of #291
Reworked #291

PR #200, opened 2017-05-19
Rewritten tests using ava snapshot

### LINKED ISSUE #195: Move from ava to Jest
It would be nice to replace ava with Jest since most of our tests are snapshots after all.
See test/fixtures/

tdeschryver:
Fixes #195

Notes:
Had to upgrade `xo`, otherwise there where errors using `.snapshot()`. Because of this update some extra linting rules were introduced (comments had to start with a capital letter, `gulpfile.babel.js` had a unused import error with `babel-polyfill`)

Had to upgrade `ava` settings in `package.json` otherwise the following error occured `./test/_transform.js: Unexpected token (10:8)`

Had to rewrite the `invalid-expressions` tests. It was written with a `forEach` but this gave linting errors and I didn't find any solution, so I've created 4 separate tests.


gulpfile.babel.js
```diff
@@ -3,6 +3,7 @@ import {join} from 'path'
 import {gzipSync} from 'zlib'
 
 // Packages
+// eslint-disable-next-line no-unused-vars
 import _ from 'babel-polyfill'
```
giuseppeg: Probably you can remove this import. Did you try that?

package.json
```diff
@@ -27,7 +27,14 @@
...
       "babel-polyfill"
-    ]
+    ],
+    "babel": {
+      "presets": [
+        "es2015",
+        "stage-3",
+        "react"
+      ]
+    }
```
giuseppeg: Can you explain further what happens if you don't have this? 
Probably `babel-register` needs it and can't pick it up from the top level one. 
We should avoid the repetition, if you want I can take a look.

package.json
```diff
@@ -67,7 +74,7 @@
     "stylis": "1.2.7"
   },
   "devDependencies": {
-    "ava": "0.17.0",
+    "ava": "^0.19.1",
```
giuseppeg: please pin this dependency i.e. remove the `^`

package.json
```diff
@@ -85,7 +92,7 @@
     "mz": "2.6.0",
     "react": "15.4.1",
     "react-dom": "15.4.1",
-    "xo": "0.17.1"
+    "xo": "^0.18.2"
```
giuseppeg: same here remove `^`

test/index.js
```diff
@@ -19,96 +18,112 @@ const transform = (file, opts = {}) => (
...
 
 test('throws when using `props` or constants ' +
   'defined in the closest scope', async t => {
-  [1, 2, 3, 4].forEach(i => {
-    t.throws(
-      transform(`./fixtures/invalid-expressions/${i}.js`),
-      SyntaxError
-    )
-  })
+  await t.throws(
```
giuseppeg: probably here it was complaining about the string concatenation 

```js
'throws when using `props` or constants ' +
    'defined in the closest scope'
```

Can you try to restore the old test and merge the strings?

giuseppeg:
@tdeschryver thank you man! You should rebase with master.

gulpfile.babel.js
tdeschryver: When I removed the error the build crashes [(https://travis-ci.org/zeit/styled-jsx/builds/234130358)](https://travis-ci.org/zeit/styled-jsx/builds/234130358).

package.json
tdeschryver: Without it we're getting errors at the spread operator @ `/test/_transform.js: Unexpected token (10:8)`. I didn't immediately find a solution for it, it would be nice if you can tackle this problem :).

test/index.js
tdeschryver: It has nothing to do with the string concatenation, it is because the test is `async`.
See: `[XO] Function was declared as async but doesn't use await (ava/no-async-fn-without-await)` 

I've tried some other solutions (using `await` in the `forEach`, using a `generator`, ...) but without success, this was the only solution that worked for me. Maybe you can steer me in the right direction if you have a solution?

giuseppeg: ah ok try:

```js
const promises = [1, 2, 3, 4].map(i => {
  t.throws(
    transform(`./fixtures/invalid-expressions/${i}.js`),
    SyntaxError
  )
})

await Promise.all(promises)
```

package.json
giuseppeg: if you rebase (git pull --rebase) and push I'll take a look

test/index.js
tdeschryver: works like charm!

package.json
tdeschryver: sorry first timer here, had some difficulties that's why it took this long... should be OK now tho! :)

giuseppeg: No worries man, this is open source and I am not your boss, there is no deadline :D 
Thanks again for contributing

giuseppeg: ok you can get rid of the duplicate babel configuration and just add `babel: "inherit"`

```json
"ava": {
  "require": [
    "babel-register",
    "babel-polyfill"
  ],
  "babel": "inherit"
},
```

gulpfile.babel.js
giuseppeg: fwiw you can remove this import and `babel-polyfill` anywhere and install `transform-runtime`

```bash
npm install --save-dev babel-plugin-transform-runtime
```

then add it to the babel configuration as plugin:

```json
"babel": {
  "presets": [
    "es2015",
    "stage-3",
    "react"
  ],
  "plugins": [
    "transform-runtime"
  ]
}
```

package.json
tdeschryver: seems good. how did I miss that...

gulpfile.babel.js
tdeschryver: Done. Thanks for the help!

giuseppeg:
@tdeschryver thank you again for the help <3

tdeschryver:
@giuseppeg thanks for the help along the way, it was a very nice first experience!
Just one more question if you don't mind, what do I do with the forked repo? Just delete it?

giuseppeg:
Well hopefully you will contribute again :D so you may want to keep the fork, you just need to keep it up to date which you can do like this:

```
git remote add upstream https://github.com/zeit/styled-jsx.git
```
and before starting new work

```
git pull upstream master
```

Also I recommend that next time you work in a branch rather than master. So after you've done the things above:

```
git checkout -b your-next-branch-name
```

tdeschryver:
I was planning to do some more contributions :), so I'll keep my branch in sync.
I've read the branching 'stuff' after the PR, but I'll do it definitely next time.
Thanks again for the input!

PR #204, opened 2017-05-21
Fix expressions scope check

### LINKED ISSUE #176: 'Not come from the closest scope' error when using constants
I'm getting the 'Expected var to not come from the closest scope' error even when the var it is not coming from the closest scope. It happens when the var is a index from a object that has the same name as one of the component props.
Ex:
```
const blue = { bg: 'blue', color: 'white' };
const orange = { bg: 'orange', color: 'white' };

const Button = ({ title, color }) => (
  <button className={color}>
    {title}
    <style jsx>{`
      button { border-radius: 5px; border: 1px solid; }
      .blue { background: ${blue.bg}; color: ${blue.color}; }
      .orange { background: ${orange.bg}; color: ${orange.color}; }
    `}</style>
  </button>
);

<Button title="go" color="blue" />
<Button title="cancel" color="orange" />
```

In that code, I'm getting the error when using ${blue.color}. If I change the name 'color' to something else (in the style object or the props), it work perfectly. Can this be a bug from the validator?

giuseppeg:
Fixes #176

### DIFF
--- a/src/_utils.js
+++ b/src/_utils.js
@@ -150,12 +150,14 @@ export const makeStyledJsxTag = (id, transformedCss, isTemplateLiteral) => {
 // The following visitor ensures that MemberExpressions and Identifiers
 // are not in the scope of the current Method (render) or function (Component).
 export const validateExpressionVisitor = {
-  MemberExpression(path) {
+  MemberExpression(path, scope) {
     const { node } = path
     if (
-      t.isThisExpression(node.object) &&
-      t.isIdentifier(node.property) &&
-      (node.property.name === 'props' || node.property.name === 'state')
+      (t.isIdentifier(node.property) &&
+        (t.isThisExpression(node.object) &&
+          (node.property.name === 'props' ||
+            node.property.name === 'state'))) ||
+      (t.isIdentifier(node.object) && scope.hasOwnBinding(node.object.name))
     ) {
       throw path.buildCodeFrameError(
         `Expected a constant ` +
@@ -166,6 +168,9 @@ export const validateExpressionVisitor = {
     }
   },
   Identifier(path, scope) {
+    if (t.isMemberExpression(path.parentPath)) {
+      return
+    }
     const { name } = path.node
     if (scope.hasOwnBinding(name)) {
       throw path.buildCodeFrameError(
--- a/test/__snapshots__/index.js.snap
+++ b/test/__snapshots__/index.js.snap
@@ -83,25 +83,28 @@ export default class {
 
 exports[`works with expressions in template literals 1`] = `
 "import _JSXStyle from 'styled-jsx/style';
+const darken = c => c;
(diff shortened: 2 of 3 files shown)


coluccini:
👏

lancejpollard:
I just downloaded 2beta and am getting this:

```
Module build failed: SyntaxError: Expected `id` to not come from the closest scope.
Styled JSX encourages the use of constants instead of `props` or dynamic values which are better set via inline styles or `className` toggling. See https://github.com/zeit/styled-jsx#dynamic-styles

   8 |     <div id={id}>
   9 |       <style jsx>{
> 10 | `#${id} {
     |     ^
  11 |   transform: translateZ(0);
  12 | }
```

Not sure if I am missing something. It is working in my Layout component, but not a deeply nested one.

lancejpollard:
@giuseppeg If I just do `#asdf` instead, so there's no variable, it gives this error:

```
Module build failed: SyntaxError: The Identifier `css` is either `undefined` or it is not an external StyleSheet reference i.e. it doesn't come from an `import` or `require` statement

```
var css = `#asdf { width: 10px; }`
  21 | }`
  22 |   return (
> 23 |     <div id={id}>
     |     ^
  24 |       <style jsx>{css}</style>
  25 |     </div>
  26 |   )
```

**Comment by giuseppeg** [comment]
@lancejpollard if you are trying to integrate v2 with Next you'll need to patch the babel configuration otherwise it will sill use 1.0

Here's an example https://github.com/zeit/next.js/tree/master/examples/with-styled-jsx-postcss in your case you'd have to require `styled-jsx/babel` rather thatn `styled-jsx-postcss/babel`

**Comment by giuseppeg** [comment]
fyi I would avoid these kind of stuff `var css = '#asdf { width: 10px; }'` since the interpolations are not preprocessed meaning that selectors won't be scoped or you won't get vendor prefixes

**Comment by lancejpollard** [comment]
@giuseppeg I have been doing this since I am running a custom next server with server-side routes:

```
babel-node --presets=es2015,stage-0 --plugins=transform-react-jsx,styled-jsx/babel server.js
```

Not sure if that is what you're describing, but still not working.

**Comment by lancejpollard** [comment]
@giuseppeg ah nice, just upgraded to `next@beta` and it works, with React 16 and everything.

**Comment by lancejpollard** [comment]
Actually, it works without using variables in the CSS, but once I add that `${id}` or whatever into the CSS it gives:

```
Module build failed: SyntaxError: Expected `id` to not come from the closest scope.
Styled JSX encourages the use of constants instead of `props` or dynamic values which are better set via inline styles or `className` toggling. See https://github.com/zeit/styled-jsx#dynamic-styles
```

giuseppeg:
@lancejpollard do you mind sharing a sample repo? like a very minimal and simplified testcase that I can try

Issue #146, opened 2017-03-08
`:not()` does not seem to be adding prefix


cgood92:
I have the following line:

```
.root > :global(*):not(header) {
...
}
```

Which get's rendered (after the babel transform) to 

```
.root[data-jsx=\"2663302060\"] > *:not(header) {
...
}
```

I guess I would expect the stuff inside the "NOT" to have prefixes as well, unless I stated that they were global.  So, what is being done and I what I would expect are different.

Here is what I would expect:

```
/* with prefix */
.root > :global(*):not(header)
=> 
.root[data-jsx=\"2663302060\"] > *:not(header[data-jsx=\"2663302060\"])
```

```
/* with global */
.root > :global(*):not(:global(header))
=> 
.root[data-jsx=\"2663302060\"] > *:not(header)
```

Styled-jsx version 0.5.7, as per dependency of next version 2.0.0-beta.32

thysultan:
Maybe [L57](https://github.com/zeit/styled-jsx/blob/master/lib/style-transform.js#L57) can also `.replace(/&/g, prefix)` to allow for & in :global() i.e 

```scss
// will allow this pattern
.root > :global(*):not(header) {}

// to be written as
.root > :global(*:not(&)) {}
```

giuseppeg:
I think that we shouldn't introduce the `&` symbol. 

Regexp are not good to match parenthesis therefore I think that we should do regular parsing like in my rewrite wip branch.

https://git.io/vyzFg
https://git.io/vyzFr
https://git.io/vyzF6

thysultan:
@giuseppeg re: parenthesis, I think i found a bulletproof solution [https://regex101.com/r/5ssplS/1](https://regex101.com/r/5ssplS/1) `:global\(((?:\[.*\]|[^()]*|\([^()]*\))*)\)`

giuseppeg:
@thysultan we could use a module like this https://github.com/juliangruber/balanced-match

thysultan:
@giuseppeg I'm using regex

```js
/:global\(((?:[^\(\)\[\]]*|\[.*\]|\([^\(\)]*\))*)\)/g
```

in stylis, it hasn't failed any of the tests i have put it through.

giuseppeg:
related #158

giuseppeg:
@thysultan any chance that you can help us out fixing this one?

Here are some test cases that should pass:

```css
p:hover {
  color: red;
}

p::before {
  color: red;
}

:hover {
[...]

PR #210, opened 2017-05-27
Upgrade to Stylis v3

### LINKED ISSUE #158: Join :global selectors
Right now this is supported:

```css
.List :global(>) :global(*) {
  margin-top: 10px;
}

.List :global(*) :global(+) :global(*) {
  margin-top: 10px;
}
```

and this is not supported:

```css
.List :global(> *) {
  margin-top: 10px;
}

.List :global(* + *) {
  margin-top: 10px;
}
```

Can we make it any better?

giuseppeg:
Fixes #158 #146


giuseppeg:
@thysultan would you have time to review this pr?

lib/style-transform.js
```diff
@@ -1,156 +1,82 @@
...
- *
- * matches :global($1) capturing $1
- */
-const regExpReplaceGlobal = /:global\((.*)\)/g
+function disableNestingPlugin() {
+  const context = arguments[0]
+  if (context === 2) {
+    const parent = (arguments[3] || [])[0]
+    if (typeof parent === 'string' && parent.trim().length > 0) {
+      throw new Error('Unfortunately nesting is not supported by styled-jsx.')
```
thysultan: @giuseppeg Optionally this could also print line and column number, so you can get to the origin of the error in the source file.

thysultan: Other than that looks perfect 👍

lib/style-transform.js
```diff
@@ -1,156 +1,85 @@
 const stylis = require('stylis')
 
-/**
- * @type {RegExp}
- *
- * matches :global($1) capturing $1
- */
-const regExpReplaceGlobal = /:global\((.*)\)/g
+function disableNestingPlugin(...args) {
+  let [context, , , parent = [], line, column] = args
```
giuseppeg: the previous way of parsing args was wrong :)

@thysultan this should do it

thysultan: @giuseppeg Is there a reason to avoid named arguments? Babel will produce a bit more code for this [http://babeljs.io/repl/#?...](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&debug=false&experimental=false&loose=false&spec=false&code=function%20disableNestingPlugin(...args)%20%7B%0A%20%20let%20%5Bcontext%2C%20%2C%20%2C%20parent%20%3D%20%5B%5D%2C%20line%2C%20column%5D%20%3D%20args%0A%20%20if%20(context%20%3D%3D%3D%202)%20%7B%0A%20%20%20%20parent%20%3D%20parent%5B0%5D%0A%20%20%20%20if%20(typeof%20parent%20%3D%3D%3D%20'string'%20%26%26%20parent.trim().length%20%3E%200)%20%7B%0A%20%20%20%20%20%20throw%20new%20Error(%0A%20%20%20%20%20%20%20%20%60Nesting%20detected%20at%20%24%7Bline%7D%3A%24%7Bcolumn%7D.%20%60%20%2B%0A%20%20%20%20%20%20%20%20%20%20'Unfortunately%20nesting%20is%20not%20supported%20by%20styled-jsx.'%0A%20%20%20%20%20%20)%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D&playground=true)

giuseppeg: args is array like not a map/object right?

thysultan: Yes, array-like.

thysultan: Oh, by named arguments i meant `(context, content, selector, parent, line, column)` not `({parent, line, column, context})`

giuseppeg: oh yep no big deal since it runs in node which doesn't need transpilation

giuseppeg: for the record the reason is that xo (the linter) complains about too many args and unused ones

PR #217, opened 2017-05-31
Fix babel transformation issues

arunoda:
We've some issues in Next.js with the latest styled-jsx on Node4.
See: https://travis-ci.org/zeit/next.js/jobs/237810896

All we've to do is move the `lib` directory into `src`
With that it'll do the babel transformation properly.


giuseppeg:
@arunoda I think that it is failing because of the spread operator `...` 

https://github.com/zeit/styled-jsx/blob/master/lib/style-transform.js#L3-L4

Can you instead replace it with 

```js
function disableNestingPlugin() {
  let [context, , , parent = [], line, column] = arguments
```

Same here: https://github.com/zeit/styled-jsx/blob/master/lib/style-transform.js#L24

giuseppeg:
Can you also add

```
node_js:
  - "4"
  - "6"
```

to `.travis.yml` while you are on this?

giuseppeg:
also destructuring is probably not supported in node4 (i don't remember), if so unwrap the args by hand

```js
function disableNestingPlugin() {
  let context = arguments[0]
  let parent = arguments[3] || []
  let line = arguments[4]
  let column = arguments[5]
```

arunoda:
We are already transforming this code. Why do we need to worry about `...`. Let babel worry about this.

giuseppeg:
Because 

1) that lib is exposed and can be used from outside, so we are making a breaking change by moving it to another folder
2) it runs in node only so imho it doesn't make sense to transpile it just for this small thing

giuseppeg:
@arunoda my point is I don't want to move the file. But we can definitely transpile it here:

https://github.com/zeit/styled-jsx/blob/master/gulpfile.babel.js#L12-L14

arunoda:
I'm kind a busy now. Feel free to change it as needed.
I just wanted to run this on Node4. That's it.

giuseppeg:
@arunoda I fixed this in #218 because it was easier for me to push there (not sure if I could push to your branch). I ended up doing what you did here but also exposing `lib/style-transform.js` which no points to the transpiled one. Thanks for your PR man!

giuseppeg:
released `1.0.3`

arunoda:
Nice work.
Now I can merge Next.js PRs :D

PR #236, opened 2017-06-09
Fix HMR not updating style

akupila:
HMR is not working for me, the `__jsx-style-xxx` id is always one behind the one on the dom node, except for the initial load. Since it's out of sync the styling disappears on the target node. 
Steps to reproduce are simply to follow the instructions for setting up a new project and changing the something in the css. The changes are not reflected unless you refresh or 

Given this basic page:
```js
export default () =>
  <div>
    <h1 className="test">Hello</h1>
    <style jsx>{`
      .test {
        border: 3px blue solid;
      }
    `}</style>
  </div>;
```

Initially:
image

After an update to the css:
image

I've been digging into this and found that [update() is called in componentWillUpdate()](https://github.com/zeit/styled-jsx/blob/master/src/style.js#L11-L13). Since the component's props are about to be updated [c.props.styleId](https://github.com/zeit/styled-jsx/blob/master/src/style.js#L27) refers to to the previous props. Thus the `styleId` gets out of sync and breaks HMR. If the lifecycle method is changed to `componentDidUpdate` this all works fine since the `this.props` (or `c.props`) are now the new ones. 

I'm not sure if there's a downside with this approach. I didn't see any issues with it but i might be missing something.

An alternative approach would be to pass in the `styleId` from the `nextProps` param to `componentWillUpdate`. I can update the PR to this if that makes more sense.

...Or maybe i'm just missing something obvious (_how is nobody else reporting this issue?.._)

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -19,13 +19,13 @@ function diff(a, b) {
 const fromServer = new Map()
 
 function patch([added, removed]) {
-  for (const [id, c] of added) {
+  for (const [id, css] of added) {
     // Avoid duplicates from server-rendered markup
     if (!fromServer.has(id)) {
       fromServer.set(id, document.getElementById(`__jsx-style-${id}`))
     }
 
-    const tag = fromServer.get(id) || makeStyleTag(c.props.css)
+    const tag = fromServer.get(id) || makeStyleTag(css)
     tags.set(id, tag)
   }
 
--- a/src/server.js
+++ b/src/server.js
@@ -4,14 +4,14 @@ import { flush } from './style'
 export default function flushToReact() {
   const mem = flush()
   const arr = []
-  for (const [id, c] of mem) {
+  for (const [id, css] of mem) {
     arr.push(
       React.createElement('style', {
         id: `__jsx-style-${id}`,
         // Avoid warnings upon render with a key
         key: `__jsx-style-${id}`,
         dangerouslySetInnerHTML: {
-          __html: c.props.css
+          __html: css
         }
       })
     )
@@ -22,8 +22,8 @@ export default function flushToReact() {
 export function flushToHTML() {
   const mem = flush()
   let html = ''
-  for (const [id, c] of mem) {
-    html += `<style id="__jsx-style-${id}">${c.props.css}</style>`
+  for (const [id, css] of mem) {
+    html += `<style id="__jsx-style-${id}">${css}</style>`
   }
   return html
 }
--- a/src/style.js
+++ b/src/style.js
@@ -8,7 +8,7 @@ export default class extends Component {
     mount(this)
   }
 
-  componentWillUpdate() {
+  componentDidUpdate() {
     update()
   }
 
--- a/src/style.js
+++ b/src/style.js
@@ -8,8 +8,12 @@ export default class extends Component {
     mount(this)
   }
 
-  componentDidUpdate() {
-    update()
+  componentWillUpdate(nextProps) {
+    update({
+      instance: this,
+      id: nextProps.styleId,
+      css: nextProps.css
+    })
   }
 
   componentWillUnmount() {
@@ -21,16 +25,22 @@ export default class extends Component {
   }
 }
 
-function componentMap() {
+function styleMap(updated) {
   const ret = new Map()
   for (const c of components) {
-    ret.set(c.props.styleId, c)
+    if (updated && c === updated.instance) {
+      // Use styleId and css from upated component rather than reading props
+      // from the component since they haven't been updated yet
+      ret.set(updated.styleId, updated.css)
+    } else {
+      ret.set(c.props.styleId, c.props.css)
+    }
   }
   return ret
 }
 
 export function flush() {
-  const ret = componentMap()
+  const ret = styleMap()
   components = []
   return ret
 }
@@ -50,6 +60,6 @@ function unmount(component) {
   update()
 }
 
-function update() {
-  render(componentMap())
+function update(updates) {
+  render(styleMap(updates))
 }
--- a/src/style.js
+++ b/src/style.js
@@ -11,7 +11,7 @@ export default class extends Component {
   componentWillUpdate(nextProps) {
     update({
       instance: this,
-      id: nextProps.styleId,
+      styleId: nextProps.styleId,
       css: nextProps.css
     })
   }
--- a/test/__snapshots__/index.js.snap
+++ b/test/__snapshots__/index.js.snap
@@ -5,7 +5,7 @@ exports[`generates source maps 1`] = `
 export default (() => <div data-jsx={188072295}>
     <p data-jsx={188072295}>test</p>
     <p data-jsx={188072295}>woot</p>
-    <_JSXStyle styleId={188072295} css={\\"p[data-jsx=\\\\\\"188072295\\\\\\"]{color:red}\\\\n/*# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInNvdXJjZS1tYXBzLmpzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUlnQixBQUNjLFdBQUMiLCJmaWxlIjoic291cmNlLW1hcHMuanMiLCJzb3VyY2VzQ29udGVudCI6WyJleHBvcnQgZGVmYXVsdCAoKSA9PiAoXG4gIDxkaXY+XG4gICAgPHA+dGVzdDwvcD5cbiAgICA8cD53b290PC9wPlxuICAgIDxzdHlsZSBqc3g+eydwIHsgY29sb3I6IHJlZCB9J308L3N0eWxlPlxuICA8L2Rpdj5cbilcbiJdfQ== */\\\\n/*@ sourceURL=source-maps.js */\\"} />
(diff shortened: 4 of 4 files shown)


giuseppeg:
@akupila thank you for the awesome explanation and PR! 
I get the issue and it would be great if we could add a failing test to reproduce the issue.

That said I think that we should stick with `componentWillUpdate` and just find a way to update the props (maybe cloning the component?). The reason why we don't `update` on `componentDidUpdate` is to avoid flash of unstyle content I believe.

cc @Rowno hi, what do you think?

akupila:
Good point on the test. I can try to figure out a failing test case.

Regarding flash of unstyled content i think it shouldn't be too much of a problem as the initial styling is based on the server render and i at least didn't see any weird behavior when updating styling. 

I agree `componentWillUpdate` is a better place to do this though. Rather than cloning the component i think the updated `styleId` can be grabbed from the arguments passed to `componentWillUpdate`, and pass it along to `update` (rather than having `update` rely on reading props from the component). I'll explore this approach during the weekend/next week.

giuseppeg:
You'd need to update all the props not just `styledId`. Here is what you could do:

```js
componentWillUpdate(nextProps) {
   updateWithProps(this, nextProps)
}

function updateWithProps(component, props) {
  const i = components.indexOf(component)
  const withProps = React.cloneElement(component, props)

  if (i < 0) {
    components.push(withProps)
  } else {
     components.splice(i, 1, withProps)
  }
  update()
}
```

giuseppeg:
thanks for the update, @Rowno recently changed it to be a map of components (before it was of styles like you did in this PR). I would like to avoid the back and forth / switch and introduce yet another breaking change.

Could you implement what I suggested above? sorry if I made you waste time :( Also we should add a unit test, let me know if you need help with that.

akupila:
I just realized that there are some problems with this update… I'm not sure how this is a breaking change (since it's only internal stuff) but i'll try the approach with cloning to component and see if it fixes the issues

giuseppeg:
... although your solution is faster 🤔 

It is breaking since `flush` now returns something else

akupila:
Yeah this is why i was thinking that cloning wouldn't be ideal. I have the breaking test fixed locally but there's an issue where if a component has multiple styles and you change one the updates don't correctly trickle through

giuseppeg:
yea because those other components' `css` is not updated right?

akupila:
With 76c8f58c6023a4ff327d619f6d3fcbde027b47ed:

```js
export default () =>
  <div>
    <h1 className="a b">Hello</h1>
    <style jsx>{`
      .a {
        border: 10px red solid;
      }
    `}</style>
    <style jsx>{`
      .b {
        color: red;
      }
    `}</style>
  </div>;
```

If you change the styling on class `.a` it's fine but `.b` styling gets dropped. If you change `.b` the `.b` styling gets dropped too

akupila:
Interestingly with the `updateWithProps` approach only the latest styling gets applied, the other one is always dropped (in the test code above) :/

giuseppeg:
Yea because the hashes for the styles need to be recalculated and applied to the elements. HMR is hard :) and probably we can't fix it that easily - in a way I think that this is the same issue that we need to solve when we will implement dynamic styles support e.g. `div { color: ${props => props.theme.something} }`

akupila:
Turns out [the problem](https://github.com/zeit/styled-jsx/pull/236#issuecomment-307726625) with just because of a typo.

There is still a small thing with this:
If you have multiple `<style jsx>` tags in a component and you update one only the latest one gets applied (and the other one removed). If there's an inline style and an external one they both work fine. I'm not sure if multiple inline styles worked before though either for HMR

akupila:
I can't think of a good way to have a test for this. @giuseppeg if you have thoughts i would love to hear them :)

In terms changing it to a `styleId` to `css` map i still think it makes more sense this way, unless i'm missing something. https://github.com/zeit/styled-jsx/pull/187 changed this but the more important change there is the change to a `Map` from a regular object. Looks like this change brought up similar attention the first time too (https://github.com/zeit/styled-jsx/pull/187#discussion_r115324161) ;)

akupila:
An alternative approach is the `cloneElement` approach. This makes the diff much smaller and is closer to the current implementation (while fixing HMR):

```diff
diff --git a/src/style.js b/src/style.js
index b82d037..a63a3ee 100644
--- a/src/style.js
+++ b/src/style.js
@@ -1,4 +1,4 @@
-import { Component } from 'react'
+import React, { Component } from 'react'
 import render from './render'

 let components = []
@@ -8,8 +8,8 @@ export default class extends Component {
     mount(this)
   }

-  componentWillUpdate() {
-    update()
+  componentWillUpdate(nextProps) {
+    updateWithProps(this, nextProps)
   }

   componentWillUnmount() {
@@ -50,6 +50,18 @@ function unmount(component) {
   update()
 }

+function updateWithProps(component, props) {
+  const i = components.indexOf(component)
+  const withProps = React.cloneElement(component, props)
+
+  if (i < 0) {
+    components.push(withProps)
+  } else {
+    components.splice(i, 1, withProps)
+  }
+  update()
+}
+
 function update() {
   render(componentMap())
 }
```

This has the same problem as my implementation where multiple `<style jsx>` tags with inline styled (= not externally imported) styles only update the latest one. However, i think this is because the `data-jsx` gets updated and only `data-jsx-ext` has the `~` check (which allows multiple data tags, similar to multiple classes). That's probably a separate PR though. 

I still think cloning the component is a bit overkill for quite a simple problem but I can replace the implementation in this PR with the one here if that seems like a better idea

arunoda:
Guys, shall we take this ASAP.
Current stable release of Next.js is breaking due to this bug.

I think this is a great solution.

giuseppeg:
> There is still a small thing with this:
If you have multiple <style jsx> tags in a component and you update one only the latest one gets applied.

> This has the same problem as my implementation where multiple <style jsx> tags with inline styled (= not externally imported) styles only update the latest one. 

@akupila do you mean that if you have

```jsx
export default () => (
  <div>
    <style jsx>{`div { color: red }`}</style>
    <style jsx>{`div { font-size: 10px }`}</style>
  </div>
)
```

and you make a change to any of them, only the last one (with `font-size`) is updated/replaced?

akupila:
Consider the following:

```jsx
export default () => (
  <div>
    <h1 className="a b">Hello</h1>
    <style jsx>{`.a { color: red }`}</style>
    <style jsx>{`.b { font-size: 10px }`}</style>
  </div>
)
```

With the above:
1. initially both styles are applied
2. change the `.a` color to blue: text becomes blue but default size
3. change the `.b` size to 64px: text becomes bigger but resets to black
4. add a `<p />` after `<h1 />`: both styles are correctly applied and the text is 64px & blue

If you put everything into a single `<style jsx />` this is not a problem. Since ~100% of examples i've seen do so anyway maybe this is not too big of a deal. Figured i'd bring this caveat up anyway since i noticed it

Regardless i think this is unrelated to the PR but you guys are more familiar with the code than I am 😉

giuseppeg:
@akupila thanks for clarifying. I'll look into this issue later or over the weekend. Can you grant me push access to your branch?

akupila:
I have this checked:
image

My understanding is that this should allow you to push commits to my branch?

giuseppeg:
@arunoda @rauchg maybe someone from the zeit team can take a stab at this. I might not have time to fix it immediately.

giuseppeg:
@akupila I wasn't able to push to your fork, https://github.com/zeit/styled-jsx/pull/245 is the failing test, with your fix it should work. Now that you have an example test can you tell me how to reproduce the other issue? I'll add a test for that too and then perpare a fix for both (possibly just use yours).

giuseppeg:
@akupila I added the test and your fix in #245 and merged. 
Thanks so much for your help man!

I was also able to reproduce the other issue which seemed to be an issue on `styled-jsx@0.5.7` so there was no regression there. I am not an expert when it comes to HMR but it seems that only the modified style tag (`JSXStyle` component when it is compiled) is replaced together with the rest of the component except for the other style tags (`JSXStyle`). I wonder if there is anything that we can do from the styled-jsx side cc @rauchg @arunoda or this has to be fixed in Next.

giuseppeg:
released `1.0.6` cc @arunoda

akupila:
Great! Happy to see this fixed and merged 🎉

arunoda:
Thanks all for your hard works.
Here are the releases of Next.js with fixes:

* stable - [2.4.5](https://github.com/zeit/next.js/releases/tag/2.4.5)
* beta -

jamilabreu:
@arunoda i'm getting a strange error from Apollo during hot reload that I wasn't getting before this update. The steps are:
1) Update jsx style
2) Error below appears
3) Refresh page manually, everything works (including new styles)

> The operation 'deletePost' wrapping 'Channels' is expecting a variable: 'id' but it was not found in the props passed to 'Apollo(Channels)'
Invariant Violation: The operation 'deletePost' wrapping 'Channels' is expecting a variable: 'id' but it was not found in the props passed to 'Apollo(Channels)'
    at invariant (http://localhost:3000/_next/1498504329264/commons.js:43155:15)
    at GraphQL.calculateOptions (http://localhost:3000/_next/1498504329264/commons.js:43435:21)
    at GraphQL.updateQuery (http://localhost:3000/_next/1498504329264/commons.js:43483:33)
    at GraphQL.componentWillReceiveProps (http://localhost:3000/_next/1498504329264/commons.js:43373:26)
    at http://localhost:3000/_next/1498504329264/commons.js:33012:23
    at measureLifeCyclePerf (http://localhost:3000/_next/1498504329264/commons.js:32477:12)
    at ReactCompositeComponentWrapper.updateComponent (http://localhost:3000/_next/1498504329264/commons.js:33011:9)
    at ReactCompositeComponentWrapper.receiveComponent (http://localhost:3000/_next/1498504329264/commons.js:32948:10)
    at Object.receiveComponent (http://localhost:3000/_next/1498504329264/commons.js:2616:22)
    at ReactCompositeComponentWrapper._updateRenderedComponent (http://localhost:3000/_next/1498504329264/commons.js:33155:23)

Issue #209, opened 2017-05-26
Add uniqueClass generator utilily


giuseppeg:
Parent components could style children by passing a classname that they control e.g.

```jsx
import { uniqueClass } from 'styled-jsx/utils'
const buttonClass = uniqueClass()

export default () => (
  <div>
      <Button className={buttonClass} />
      <style jsx>{`
          .${buttonClass} { color: red }
      `}</style>
  </div>
)
```

In order to do so we'd need to transpile the import to something like

```js
const uniqueClass = (() => {
   const uid = hash(thisModuleRelativePath)
   const counter = 0
   return () => `__style-jsx-${uid}-${counter++}`
})()
```

Where `thisModuleRelativePath` could be something like [this](https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/visitors/displayNameAndId.js#L9-L13).

@nkzawa @rauchg thoughts? Do you think that this could be useful?

pungggi:
Hi any news on that?

giuseppeg:
Found a better mouse trap: https://github.com/zeit/styled-jsx#styling-third-parties--child-components-from-the-parent

PR #244, opened 2017-06-23
allow targeting root with dynamic tag name

### LINKED ISSUE #147: Can't target root with a dynamic tag name
You can't [target a root element](https://github.com/zeit/styled-jsx#targeting-the-root) with a [dynamic tag name](http://stackoverflow.com/a/33471928/1596978), using `className`:

```js
export default ({children, level = 2}) => {
  const Heading = `h${level}`
  return (
    <Heading className='root'>
      {children}
      <style jsx>{`
        .root {
          color: red;
        }
      `}</style>
    </Heading>
  )
}
```

SomeHats:
Thanks for styled-jsx! This should fix #147.

Just as a note: lint-staged was trying to lint the fixture I had which it probably shouldn't do - none of the other fixtures will lint, but as I was staging that file it was getting upset.

I worked around it and committed without linting in the end because I'm some sort of git criminal I guess

Let me know if you need any changes :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "styled-jsx",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "styled-jsx",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "license": "MIT",
   "repository": "zeit/styled-jsx",
   "description": "Full CSS support for JSX without compromises",


giuseppeg:
aw thank you for the PR! 

Just a nit / consideration, let me know what you think about it:

This patch allows people to define components within the render method
which yeah.. I realize that is not that common to have but I've seen that. Do you think that it'd be hard and/or a good idea to check whether the binding is just a string/template literal?

In terms of tests this means:

```jsx
// valid

() => {
   const Component = 'h1' 
}

// not valid

() => {
   const Component = () => <div />
}
```

And implementation wise it could be something like this http://astexplorer.net/#/gist/2625ed29e3b13bf34f5bab17f0deb6ea/a27e6bc3fc6e70bda8d205efe84990a76f6fd8af you can then use `state.validComponents` (maybe find a better name) to check whether the component is whitelisted / legit.

I am asking because I believe that if users currently have components within the render method and upgrade then their tests (snapshots) might fail.

> Just as a note: lint-staged was trying to lint the fixture I had which it probably shouldn't do - none of the other fixtures will lint, but as I was staging that file it was getting upset.

yeah no worries! maybe we shouldn't lint fixtures :)

SomeHats:
Hmm that's a good point. How about some sort of explicit opt-in syntax for components instead? `<Component data-jsx />` or something.

I guess this is sort of reflecting a wider issue around component composition with styled-jsx. These two components are styled-jsx versions of ones I write fairly often:
```js
const Button = ({ as = 'button', children, ...props }) => {
  const Component = as;
  return (
    <Component {...props}>
      {children}

      <style jsx>{`
        .root {
          /* base button styles */
        }
      `}</style>
    </Component>
  );
};

const LoadingButton = ({ loading, children, ...props }) => (
  <Button {...props}>
    {loading
      ? 'Loading...'
      : children}

    <style jsx>{`
      .root {
        /* override/extend Button's styles */
      }
    `}</style>
  </Button>
);
```

This PR makes `Button` possible - it lets me do `<Button as="a" />` and still have my styles applied. But the second one (afaict) isn't really possible with styled-jsx. We'd need some way of making sure every data-jsx value gets passed down and then have several values on the dom elements, which would be pretty tricky.

Anyway this discussion is outside the scope of this PR, but is this a pattern you've come across yourself? How do you (or others) currently do this sort of thing with styled-jsx?

giuseppeg:
@SomeHats as for now I think the way to go is wrapping components + `:global`. 

```jsx
const LoadingButton = ({ loading, children, ...props }) => (
<span>
  <Button {...props}>
    {loading
      ? 'Loading...'
      : children}
  </Button>
    <style jsx>{`
      span > :global(.root) {
        /* override/extend Button's styles */
      }
    `}</style>
</span>
);
```
Maybe #209 could make it easier to customize a child. Ideas are welcome.

Anyway, back to this PR I think that we should do what I suggested for now.

Maybe instead of allowing string/template literals only we can allow anything but functions e.g. `const Component = () => {}` or `Identifier`s whose value starts with a capital letter eg. `const Component = Foo`.

Issue #251, opened 2017-06-27
External stylesheets break with Prettier 1.5.0


leo:
In some cases, this:

```jsx
<style jsx>{styles}</style>
```

...becomes this:

```jsx
<style jsx>
  {styles}
</style>
```

The above throws an error. Not sure if it's a Next.js issue or not.

rauchg:
What's the error?

giuseppeg:
@leo what is the stack trace?

leo:
Whoops, sorry!

```
> Using "webpack" config function defined in next.config.js.
> Failed to build on /var/folders/c0/ns546fkj6t52nvlsm2zvm6p40000gn/T/8f4b0e64-cfbf-478b-a9e7-81600f7e9c73
{ Error: ./renderer/components/feed/switcher.js
Module build failed: TypeError: /Users/leo/Projects/now-desktop-stage/renderer/components/feed/switcher.js: Invalid attempt to destructure non-iterable instance
    at /Users/leo/Projects/now-desktop-stage/node_modules/babel-runtime/helpers/slicedToArray.js:48:13
    at PluginPass.exit (/Users/leo/Projects/now-desktop-stage/node_modules/styled-jsx/dist/babel.js:250:66)
    at newFn (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/visitors.js:276:21)
    at NodePath._call (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/path/context.js:48:17)
    at NodePath.visit (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/path/context.js:117:8)
    at TraversalContext.visitQueue (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/context.js:150:16)
    at TraversalContext.visitMultiple (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/context.js:103:17)
    at TraversalContext.visit (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/context.js:190:19)
    at Function.traverse.node (/Users/leo/Projects/now-desktop-stage/node_modules/babel-traverse/lib/index.js:114:17)
 @ ./renderer/pages/feed.js?entry 105:16-54
 @ multi ./pages/feed.js?entry
[...]

PR #255, opened 2017-07-03
Fix prettier 1.5.x

akupila:
Prettier formats the code so that

```jsx

export default () =>
  <div>
    <h1 className="test">Hello</h1>
    <style jsx>{styles}</style>
  </div>;
```

becomes

```jsx

export default () =>
  <div>
    <h1 className="test">Hello</h1>
    <style jsx>
      {styles}
    </style>
  </div>;
```

This led to a confusing error `Invalid attempt to destructure non-iterable instance`:

image

@giuseppeg [proposed a fix for this problem](https://github.com/zeit/styled-jsx/issues/251#issuecomment-311453600), this PR is just exactly that fix + a test that fails before the fix.

In addition there is another commit which bumps prettier to 1.5.2 and formats that test fixture using prettier (so that future updates to prettier should still work). Not sure if this PR should be the place for it, if not i can remove the commit

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "prepublish": "gulp transpile",
     "test": "npm run lint && ava",
     "lint": "xo",
-    "format": "prettier --single-quote --no-semi --write *.js {src,test}/*.js",
+    "format": "prettier --single-quote --no-semi --write *.js {src,test}/*.js test/fixtures/*-prettier.js",
     "precommit": "xo --quiet && lint-staged"
   },
   "ava": {
@@ -85,7 +85,7 @@
     "husky": "^0.13.3",
     "lint-staged": "^3.4.2",
     "mz": "2.6.0",
-    "prettier": "^1.3.1",
+    "prettier": "^1.5.2",
     "react": "15.4.1",
     "react-dom": "15.4.1",
     "xo": "0.18.2"
--- a/test/fixtures/external-stylesheet-prettier.js
+++ b/test/fixtures/external-stylesheet-prettier.js
@@ -1,0 +1,9 @@
+import styles from "./styles"
+
+export default () =>
+  <div>
+    <p>test</p>
+    <style jsx>
+      {styles}
+    </style>
+  </div>
--- a/test/index.js
+++ b/test/index.js
@@ -93,6 +93,11 @@ test('works with external stylesheets', async t => {
   t.snapshot(code)
 })
 
+test('works with external stylesheets (prettier)', async t => {
+  const { code } = await transform('./fixtures/external-stylesheet-prettier.js')
+  t.snapshot(code)
+})
+
 test('works with external stylesheets (global only)', async t => {
   const { code } = await transform('./fixtures/external-stylesheet-global.js')
   t.snapshot(code)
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "prepublish": "gulp transpile",
     "test": "npm run lint && ava",
     "lint": "xo",
-    "format": "prettier --single-quote --no-semi --write *.js {src,test}/*.js",
+    "format": "prettier --single-quote --no-semi --write *.js {src,test}/*.js test/fixtures/*-prettier.js",
```
giuseppeg: i don't think that we need to format the fixture, the bug can be reproduced without prettier. can you revert this change?

test/index.js
```diff
@@ -93,6 +93,11 @@ test('works with external stylesheets', async t => {
   t.snapshot(code)
 })
 
+test('works with external stylesheets (prettier)', async t => {
+  const { code } = await transform('./fixtures/external-stylesheet-prettier.js')
+  t.snapshot(code)
```
giuseppeg: I guess that you can merge this with the existing test for external styles. If you want to keep them separate then maybe we should rename it because as I said early the bug it is not about prettier

test/fixtures/external-stylesheet-prettier.js
```diff
@@ -0,0 +1,9 @@
+import styles from "./styles"
+
+export default () =>
+  <div>
+    <p>test</p>
+    <style jsx>
+      {styles}
+    </style>
```
giuseppeg: can you add a few test cases more?

```jsx
<style jsx>{

styles}
</style>
<style jsx>{

styles
}
</style>
<style jsx>

{

styles}
</style>
<style jsx>{styles}
</style>
<style jsx>

{styles}</style>
<style jsx>


{styles}

</style>

<style jsx>{'div { color: red}'}

{styles}
</style>
```

leo:
Thank you very much, @akupila! I'll make sure to mention you in the GitHub release.

However, we needed this fixed very quickly, so I had to introduce the changes by myself in https://github.com/zeit/styled-jsx/pull/256. I hope you can understand that! 😊

package.json
akupila: The thinking was that if a prettier 1.6 comes out we want to ensure it doesn't break the babel plugin parsing. A bit of an edge case i guess

test/fixtures/external-stylesheet-prettier.js
akupila: Oh, yeah. Should have done that but was in a hurry :)

akupila:
Damn, missed the notification about responses and was too slow to respond :) That's great though, good to see this merged 🎉

akupila:
I think the additional tests @giuseppeg pointed out would have been good to add though before merging

giuseppeg:
thank you guys for taking this, i am on holiday and without a laptop

leo:
@akupila I don't have time right now and we need the fix to be released (will do in a sec). But feel free to do another PR, if you want 😊

package.json
giuseppeg: the thing is that @leo found this bug because of prettier but the formatting which caused the issue can also be done by hand

akupila: True. I think formatting a fixture as part of the the `format` is also something i don't really like.. -> Totally ok to have it removed :)

Issue #42, opened 2016-12-18
Support for nested <style jsx>


rauchg:
I haven't had the need myself to do this, but it might be interesting to be able to sub-scope:

```js
<div>
  {
    stuff.map((child) => (
      <div>
         <style jsx /> /* scoped to sub-div */
      </div>
    ))
  }
  <style jsx />
</div>
```

ericf:
I ran into this limitation my first try using `styled-jsx` 😄 Here's what I tried to do:

```js
<div className='root'>
  <header className='header'>
    ...
    <style jsx>{`
      .header {}
    `}</style>
  </header>

  <style jsx>{`
    .root {}
  `}</style>
</div>
```

I wanted to group the styles related to `<header>` and its children inside the `<header>` element itself, but got a compilation error when trying to do so.

giuseppeg:
This looks like a single component to me why would you group styles like that instead of using selectors?

ericf:
Because I was thinking of it as `<style scoped>`, and wanted to group styles with their relevant elements. This would also make it quicker to refactor later. The situation I had here was an entire document with a few main sections and my first thought was that I could co-locate my styles within the element tree.

rauchg:
Use selectors, or separate into a different component.

Not supporting that might actually end up encouraging good practices 🤔

giuseppeg:
yup also the following already works:

```jsx
export default class {
  renderHeader() {
    return (
      <header className='banner'>
        ...
        <style jsx>{`
          .banner {}
        `}</style>
      </header>
    )
  }

  render () {
    return (
      <div className='root'>
        {this.renderHeader()}
        <style jsx>{`
          .root {}
        `}</style>
      </div>
    )
  }
}
```

rauchg:
Very cool. Great example @giuseppeg

ericf:
I do like the restriction of not supporting nested `<style jsx>` contexts within a JSX syntax tree, and 👍 on the example of the style scoping working when separate JSX syntax trees are used and composed together via render functions.

I think where this lands is probably a better error message for the example I posted above. I got an error about trying to iterate on a non-iterable iirc.

rauchg:
I was about to say the same. The key is having really good error messages (as always)

katopz:
Refer to https://github.com/zeit/styled-jsx/issues/50#issuecomment-269089363
[...]

PR #260, opened 2017-07-08
Fix nested style tags break compilation

giuseppeg:
iirc we don't allow nested style tags like:

```jsx
<div>
  <span>
      <style jsx>{...}</style>
  </span>
</div>
```

With this PR we ignore them because otherwise compilation breaks when there is one.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,9 +71,9 @@
       }
     },
     "@types/node": {
-      "version": "8.0.53",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-8.0.53.tgz",
-      "integrity": "sha512-54Dm6NwYeiSQmRB1BLXKr5GELi0wFapR1npi8bnZhEcu84d/yQKqnwwXQ56hZ0RUbTG6L5nqDZaN3dgByQXQRQ==",
+      "version": "8.5.1",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-8.5.1.tgz",
+      "integrity": "sha512-SrmAO+NhnsuG/6TychSl2VdxBZiw/d6V+8j+DFo8O3PwFi+QeYXWHhAw+b170aSc6zYab6/PjEWRZHIDN9mNUw==",
       "dev": true
     },
     "abab": {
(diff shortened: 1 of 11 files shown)


giuseppeg:
@rauchg should we instead throw an error when we detect nested `style` tags?

nkzawa:
I think we should support this since scoped `style` tag works like that (as far as I know).

giuseppeg:
@nkzawa for the record I found the issue https://github.com/zeit/styled-jsx/issues/42 w

giuseppeg:
@nkzawa , @rauchg can I merge this?

rauchg:
I don't think ignoring is the right solution. Throwing an error until we fully support it would be ideal

a-ignatov-parc:
Any news on this PR? It's been a while since last activity :)

andrechalella:
Hi, first of all thanks for the great product!

I just lost a couple hours of work (and I'm on a tight schedule) because Next.js threw up with a cryptic error when I made a simple nested style.

I'd been searching the web for more than 1 hour when I finally understood (from several issues/bug reports) that styled-jsx does not allow this kind of nesting.

It was frustrating to learn that, instead of throwing early and clearly, the team decided to simply ignore invalid code.

> With this PR we ignore them because otherwise compilation breaks when there is one.

It seems obvious to me that compilation should not succeed when there is invalid code.

**Could someone explain the rationale behind this?** Maybe no one thought that users might be confused, was that the case? In case there's no good explanation, maybe this decision could be reversed, so styled-jsx would throw with a clear message (as per @rauchg's [comment](https://github.com/vercel/styled-jsx/pull/260#issuecomment-320471286))?

Also, nowhere in the docs I couldn't find that we weren't supposed to do such nesting. Maybe that needs fixing too?

---

I'll leave some details below for people that might run into the same problem as I did.

Error:

```
wait  - compiling...
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', crates/styled_jsx/src/lib.rs:531:44
error - ./pages/test.js
Error: failed to process
```

pages/test.js:

```
export default function Home() {
  return (
    <div>
      blue
      <p>
        red
        <style jsx>{` p { color: red; } `}</style>
      </p>
      <style jsx>{` div { color: blue; } `}</style>
    </div>
  )
}
```

I was able to solve it with `resolve`. I don't like defining fixed class names, because they might clash with the global styles.

```
import css from 'styled-jsx/css';

const { className, styles } = css.resolve`
  p { color: red }                                                                                  
`;

export default function Home() {
  return (
    <div>
      blue
      <p className={className}>red</p>
      <style jsx>{` div { color: blue; } `}</style>
      {styles}
    </div>
  )
}
```

PR #263, opened 2017-07-12
Fix issue with undefined this.hub.file

eliot-akira:
Fixes zeit/next.js#2531.

Not totally satisfied with the fix, since it doesn't address what's causing `this.hub` to be undefined,  upstream in the build pipeline of Next.js.

In the issue discussion, another solution was suggested using MemberExpression instead of AssignmentExpression to check for `module.exports`, but I couldn't get one of the tests to pass.

This current pull request both solves the issue and passes all tests.

### DIFF
--- a/src/babel-external.js
+++ b/src/babel-external.js
@@ -144,7 +144,8 @@ export const namedExportDeclarationVisitor = (path, opts) => {
 }
 
 export const moduleExportsVisitor = (path, opts) => {
-  if (path.get('left').getSource() !== 'module.exports') {
+  const left = path.get('left')
+  if (!left.hub || left.getSource() !== 'module.exports') {
     return
   }
   defaultExports(path, path.get('right'), opts)


giuseppeg:
I would like to figure out what's is the real cause of this bug and come up with a proper fix if you don't mind.
I just need a bit of time cos I have a full time job so I cannot jump to it immediately :)

AFAIK one of the reason why `this.hub` is not be defined might be that at some point the file is read in node (eg. `readFileSync`) and then transformed by [`babel.transform`](https://babeljs.io/docs/usage/api/#babel-transform-code-string-options-object-). I think that going for the `MemberExpression` way should be more reliable but I need to try why it failed for you :)

eliot-akira:
Sure, I agree that it would be better to address the cause of the issue in the build process, rather than "patch" it specifically in styled-jsx.

I hope the [example repo that reproduces the error](https://github.com/eliot-akira/next-build-debug) would be helpful in tracking it down. It's strange but it must have the right combination of characters that trigger the issue. I've cloned it and confirmed that the build error occurs every time.

giuseppeg:
@eliot-akira discarding this in favor of #263. Thanks so much for your help man!

PR #268, opened 2017-07-20
Upgrade stylis

### LINKED ISSUE #261: Pseudo-element selectors break compiled output
On the latest version 1.0.8:

When rendered in a react element:

    <style jsx global>{`
		.other-selector {
		  display: none;
		}

		.corrupting-selector:before
		{
		}
    `}</style>

I would expect to see:

    .other-selector {display: none}

I actually see

    <style id="__jsx-style-11968641476">{.corrupting-selector:before;{;}.other-selector{display:none}</style>

Which is invalid CSS, causing other styles defined in that block to be ignored.

I can replicate this by adding that code to one of the test fixtures and running the unit tests as well.

giuseppeg:
Fixes #261


src/lib/style-transform.js
```diff
@@ -1,4 +1,6 @@
-const stylis = require('stylis')
+const Stylis = require('stylis')
+
+const stylis = new Stylis()
```
giuseppeg: @thysultan looks good?

thysultan: 👍  LGTM.

Issue #265, opened 2017-07-15
React Fiber v16.0.0-alpha.13 support


msand:
My pull request https://github.com/zeit/next.js/pull/2570 for Fiber support in next.js fails because of peerinvalid Peer styled-jsx@1.0.8 wants react@15.x.x
Is anyone looking into Fiber compatibility?

giuseppeg:
styled-jsx is not fiber ready and we would need some refactoring to support it.
Would you like to explore a way to make it work?

dsifford:
@giuseppeg I'd love to help with this if still needed.

Curious what the exact issue is for this? I just pulled the repo, installed `react@next` and `react-dom@next` and did nothing else and all the tests pass?

Is there something I'm missing?

dsifford:
@giuseppeg Just took it a step further and dropped my locally updated `styled-jsx` with react and react-dom 16 installed into one of my projects and it appears that everything is working as expected.... I can't get it to _not_ work!

Did the React team fix whatever the problem was on their end perhaps?

callumlocke:
For me, styled-jsx@2.0.0-beta.1 seems to be working perfectly with React@16.0.0 (despite peer deps warning during npm install).

giuseppeg:
@callumlocke yea but it should fail as soon as you try to return an array of react elements I think

PR #284, opened 2017-09-17
upgrade to react 16 rc

dsifford:
Hello,

It was my intention to sit down and try to work through whatever issue was preventing styled-jsx from being able to upgrade to React 16. However, after simply cloning the repo and installing both `react@next` and `react-dom@next`, all the tests pass without me having to do anything.

What's more, I wanted to test to make sure there wasn't some other uncovered issue in the codebase by physically trying it in one of my projects and (as described in #265) I can't get this to _not_ work. 

It seems that everything is working fine.

.....Full stop.

HOWEVER! If there is something I am missing and it in fact does not work in some other scenario that I haven't tested, I'm more than happy to work on taking care of it. 

Please let me know! Thanks in advance :smile:

### DIFF
--- a/src/_utils.js
+++ b/src/_utils.js
@@ -95,6 +95,12 @@ export const isStyledJsx = ({ node: el }) =>
   el.openingElement.attributes.some(attr => attr.name.name === STYLE_ATTRIBUTE)
 
 export const findStyles = path => {
+  if (Array.isArray(path.container)) {
+    return [...path.parentPath.get('elements'), ...path.get('children')].filter(
+      isStyledJsx
+    )
+  }
+
   if (isStyledJsx(path)) {
     const { node } = path
     return isGlobalEl(node.openingElement) ? [path] : []
--- a/test/__snapshots__/index.js.snap
+++ b/test/__snapshots__/index.js.snap
@@ -71,48 +71,48 @@ exports[`works with array children 1`] = `
 "import _JSXStyle from \\"styled-jsx/style\\";
 /* eslint-disable no-unused-vars */
 
-const Test1 = () => [<div data-jsx={465566775}>Foo</div>, <span data-jsx={465566775}>Bar</span>, <_JSXStyle styleId={465566775} css={\\"div[data-jsx=\\\\\\"465566775\\\\\\"]{color:red}\\"} />];
+const Test1 = () => [<div data-jsx={1486255031}>Foo</div>, <span data-jsx={1486255031}>Bar</span>, <_JSXStyle styleId={1486255031} css={\\"div[data-jsx=\\\\\\"1486255031\\\\\\"]{color:red}\\"} />];
 
 const Test2 = () => [<p data-jsx={2639504339}>Hello World</p>, <_JSXStyle styleId={2639504339} css={\\"p{color:white}\\"} />];
 
-const Test3 = () => [<div data-jsx={2293543665}>
-        <p data-jsx={2293543665}>Hello World</p>
-        <_JSXStyle styleId={2293543665} css={\\"p[data-jsx=\\\\\\"2293543665\\\\\\"]{color:blue}\\"} />
-    </div>, <div>Hello World</div>];
+const Test3 = () => [<div data-jsx={4193872369}>
+    <p data-jsx={4193872369}>Hello World</p>
+    <_JSXStyle styleId={4193872369} css={\\"p[data-jsx=\\\\\\"4193872369\\\\\\"]{color:blue}\\"} />
+  </div>, <div>Hello World</div>];
 
-const Test4 = () => [<div data-jsx={1371748502}>
-        <span data-jsx={1371748502}>Styled by root style tag</span>
-        <p className=\\"one\\" data-jsx={1371748502}>Lorem</p>
-        <p className=\\"two\\" data-jsx={1371748502}>Ipsum</p>
-        <_JSXStyle styleId={3751795765} css={\\"span[data-jsx=\\\\\\"1371748502\\\\\\"]{background:magenta}\\"} />
-    </div>, <div data-jsx={3751795765}>
-        <p className=\\"one\\" data-jsx={3751795765}>Dolor</p>
-        <p className=\\"two\\" data-jsx={3751795765}>Sit</p>
-    </div>, <div data-jsx={1844802545}>
-        <p className=\\"one\\" data-jsx={1844802545}>Dolor</p>
-        <p className=\\"two\\" data-jsx={1844802545}>Sit</p>
-        <_JSXStyle styleId={3751795765} css={\\"span{background:magenta}\\"} />
-    </div>, <_JSXStyle styleId={3751795765} css={\\"span[data-jsx=\\\\\\"3751795765\\\\\\"]{background:magenta}\\"} />];
+const Test4 = () => [<div data-jsx={493673878}>
+    <span data-jsx={493673878}>Styled by root style tag</span>
+    <p className=\\"one\\" data-jsx={493673878}>Lorem</p>
+    <p className=\\"two\\" data-jsx={493673878}>Ipsum</p>
+    <_JSXStyle styleId={1597461941} css={\\"span[data-jsx=\\\\\\"493673878\\\\\\"]{background:magenta}\\"} />
+  </div>, <div data-jsx={1597461941}>
+    <p className=\\"one\\" data-jsx={1597461941}>Dolor</p>
+    <p className=\\"two\\" data-jsx={1597461941}>Sit</p>
+  </div>, <div data-jsx={4046372977}>
+    <p className=\\"one\\" data-jsx={4046372977}>Dolor</p>
+    <p className=\\"two\\" data-jsx={4046372977}>Sit</p>
+    <_JSXStyle styleId={1597461941} css={\\"span{background:magenta}\\"} />
+  </div>, <_JSXStyle styleId={1597461941} css={\\"span[data-jsx=\\\\\\"1597461941\\\\\\"]{background:magenta}\\"} />];
 
-const Test5 = () => [<div data-jsx={387450454}>Foo</div>, <div data-jsx={387450454}>
-        <div data-jsx={387450454}>
-            <div className=\\"bar\\" data-jsx={387450454}>Bar</div>
-        </div>
-    </div>, <div data-jsx={387450454}>
-        <main data-jsx={387450454}>
-            <div data-jsx={387450454}>
-                <div id=\\"baz\\" data-jsx={387450454}>Baz</div>
-            </div>
-        </main>
-    </div>, <_JSXStyle styleId={387450454} css={\\"main[data-jsx=\\\\\\"387450454\\\\\\"]{color:red}.bar[data-jsx=\\\\\\"387450454\\\\\\"]{color:white}.baz[data-jsx=\\\\\\"387450454\\\\\\"]{color:blue}\\"} />];
+const Test5 = () => [<div data-jsx={3832907478}>Foo</div>, <div data-jsx={3832907478}>
+    <div data-jsx={3832907478}>
+      <div className=\\"bar\\" data-jsx={3832907478}>Bar</div>
+    </div>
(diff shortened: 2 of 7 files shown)


giuseppeg:
@dsifford thanks! What won't work are components that return an array of elements e.g.
```jsx
() => [
  <div>hi</div>,
  <p>there</p>,
  <style jsx>{`div { color: red }`}</style>
]
```
styled-jsx was not written with that model in mind so it can't transpile them.
If we want to enable react 16 we should at least throw an error when styled-jsx is used like that until we support arrays.

dsifford:
Aha... Got it. That makes sense.

I'll take a peek at this some more later this week and see if I can sort something out or at the very least add the error.

dsifford:
@giuseppeg I think I have it either working or _really close_ to working with array children. Just pushed my changes for your review.

Standby for some inline comments for the specifics regarding what I mean by "really close to working".....

test/__snapshots__/index.js.snap
```diff
@@ -67,6 +67,54 @@ const Test = () => <div data-jsx={1535297024}>
...
+const Test1 = () => [<div data-jsx={465566775}>Foo</div>, <span data-jsx={465566775}>Bar</span>, <_JSXStyle styleId={465566775} css={\\"div[data-jsx=\\\\\\"465566775\\\\\\"]{color:red}\\"} />];
+
+const Test2 = () => [<p data-jsx={2639504339}>Hello World</p>, <_JSXStyle styleId={2639504339} css={\\"p{color:white}\\"} />];
+
+const Test3 = () => [<div data-jsx={2293543665}>
+        <p data-jsx={2293543665}>Hello World</p>
+        <_JSXStyle styleId={2293543665} css={\\"p[data-jsx=\\\\\\"2293543665\\\\\\"]{color:blue}\\"} />
+    </div>, <div>Hello World</div>];
+
+const Test4 = () => [<div data-jsx={1371748502}>
```
dsifford: This case is the one that's giving me the most trouble.

I'm not sure what the desired behavior would be in this situation but, as you can see, if there is a div element with a nested `<style>` component **and** a top-level `<style>` component, the top level component wins.

IMHO, this situation should merge the nested style rules with the top level style rules in each of the nested style components and keep the style rules as-is in the top-level component. 

Any ideas?

test/__snapshots__/index.js.snap
```diff
@@ -67,6 +67,54 @@ const Test = () => <div data-jsx={1535297024}>
...
+const Test1 = () => [<div data-jsx={1486255031}>Foo</div>, <span data-jsx={1486255031}>Bar</span>, <_JSXStyle styleId={1486255031} css={\\"div[data-jsx=\\\\\\"1486255031\\\\\\"]{color:red}\\"} />];
+
+const Test2 = () => [<p data-jsx={2639504339}>Hello World</p>, <_JSXStyle styleId={2639504339} css={\\"p{color:white}\\"} />];
+
+const Test3 = () => [<div data-jsx={4193872369}>
+    <p data-jsx={4193872369}>Hello World</p>
+    <_JSXStyle styleId={4193872369} css={\\"p[data-jsx=\\\\\\"4193872369\\\\\\"]{color:blue}\\"} />
+  </div>, <div>Hello World</div>];
+
+const Test4 = () => [<div data-jsx={493673878}>
```
dsifford: This case is the one that's giving me the most trouble.

I'm not sure what the desired behavior would be in this situation but, as you can see, if there is a div element with a nested `<style>` component **and** a top-level `<style>` component, the top level component wins.

IMHO, this situation should merge the nested style rules with the top level style rules in each of the nested style components and keep the style rules as-is in the top-level component. 

Any ideas?

giuseppeg: yeah the reasons why we only support top level only are

1) a subtree might not be rendered under some circumstances and that will cause styles to unmount for all the instances e.g.:

```jsx
<div>
   {this.state.isOpen && 
       <div>
          open
          <style jsx>{`div { color: red }`}</style>
       </div>
   }
</div>
```

This will be solved by dynamic styles and another feature I have in mind that will add conditions to tags using this syntax `<style jsx when={expression}>{...}</style>`, in the example `expression` would be `this.state.isOpen`.

2) a problem with nested tags is that the user might think that the styles apply only to the closest scope. In the example above to the closest `div` only not the outer one (the root).

We can try to find a solution that makes the most sense.

giuseppeg:
@dsifford thank you man for your contribution! I should have probably mention this early, I am about to finish a branch where I rewrote a lot of code to support dynamic styles (props) so before you do more work I would wait until that is done and merged and continue from there. I'll try to be quick and help you once I am done with that – hopefully I will manage to finish this weekend. Sorry for bringing this up a bit late, I think that the majority of your code will still work .. I just want to avoid crazy merge conflicts.

dsifford:
@giuseppeg No worries! I didn't make really that many changes to be honest. Most of the diff of the PR is the result of prettier reformatting. 

Happy to hold off until you get the work done on your end. 👍

test/__snapshots__/index.js.snap
dsifford: I see.

What about literally propagating the actual style text down the tree, rather than trying to do fancy stuff with the scoped selector?

**Example:** 

Given this component tree...

```js
export default () => [
  <div>
    {this.state.isOpen && (
      <div>
        open
        <style jsx>{`
          div {
            color: red;
          }
        `}</style>
      </div>
    )}
    <style>{`div { color: blue; }`}</style>
  </div>,
  <style jsx>{`
    div {
      color: white;
    }
  `}</style>
];

```

The output would render like this....

```js
export default () => [
  <div>
    {this.state.isOpen && (
      <div>
        open
        <style jsx>{`
          div {
            color: white;
          }
          div {
            color: blue;
          }
          div {
            color: red;
          }
        `}</style>
      </div>
    )}
    <style>{`
	div { color: white; }
	div { color: blue; }
   `}</style>
  </div>,
  <style jsx>{`
    div {
      color: white;
    }
  `}</style>
];
```

That way, any duplicate selectors would automatically take precedence in the child `<style>` blocks and any non-dupes would still propagate downward?

Thoughts? Am I totally leaving out something obvious that makes this a more challenging?

giuseppeg: uhm I am not sure about that. In my opinion styles like that should apply as if they were contained inside of the root element without us having to do that kind of "propagation".

I haven't tried this yet but I think that if you don't decrease `state.ignoreClosing` here https://github.com/zeit/styled-jsx/pull/284/files#diff-bb35c33015964b91f5c83deb2a33a814R263 it might work. Basically for now we shouldn't worry about nested styles, we just want the following to work:


```jsx
() => [
  <div>
      hi
      <style jsx>{`div { font-weight: bold }`}</style>
  </div>,
  <p>there</p>,
  <style jsx>{`div { color: red }`}</style>
]
```

giuseppeg: or just warn / error when `path.parentNode` of a style tag is an `ArrayExpression` :)

dsifford: You're right. Just tried to quickly write out what the process would look like if we were to propagate styles downward and it quickly became overly complex with no real benefit.

Question on your proposition:

In your example above, what would your output be?

You have a single div with conflicting style rules, one from a child style element and one from the root style element. 

If the logic proceeds as it currently does, only one of those rules will apply to the div in your example (whichever is located first in state.styles, since styles are received by using `Array.shift()`)

Consider the following situation...

```js
() => [
  <div>
    // contents....
    <style>{/* style here */}</style>
  </div>,
  <div>
    // contents....
    <style>{/* style here */}</style>
  </div>,
  <style>{/* style here */}</style>,
];
```

What would be expected to happen here?

dsifford: (Having said that, I'm totally fine with throwing the error as you mentioned. Just making sure we're on the same page.)

src/_utils.js
```diff
@@ -23,6 +23,12 @@ export const isStyledJsx = ({ node: el }) =>
   el.openingElement.attributes.some(attr => attr.name.name === STYLE_ATTRIBUTE)
 
 export const findStyles = path => {
+  if (Array.isArray(path.container)) {
+    return [...path.parentPath.get('elements'), ...path.get('children')].filter(
```
dsifford: Switching the order of these might also result in a better output

giuseppeg:
@dsifford yesterday I released styled-jsx v2 would you like to rebase or start a new branch?

dsifford:
Sure! Got some other priorities to take care of first but I'd be glad to help where I can.

The v2 release adds new features and fixes everything except the array children issue, correct?

giuseppeg:
> The v2 release adds new features and fixes everything except the array children issue, correct?

I didn't do any work related to React 16. One more thing to keep in mind (could be a followup) is to see how we can make [renderToNodeStream()](https://reactjs.org/docs/react-dom-server.html#rendertonodestream) work. 

@jpcummins 👋 you guys work together?

jpcummins:
Nope. Just working on a side project with React 16 and stylized-jsx. It’s
an awesome framework!

dsifford:
@giuseppeg Tried just now to sit down for a bit and take a stab some more at this but I can't get the current test render workflow to work with react16 due to `react-jsdom` being outdated.

Suggestions? Any way we can drop that dependency and use `react-test-renderer` instead?

giuseppeg:
@dsifford don't worry about that test for now. If necessary we can disable it but unfortunately we need the DOM since it is testing some side effects (styles being injected into the `head`).

By the way if you want to chat you can find me on Slack https://zeit.chat as `g`

dsifford:
@giuseppeg Aha, got it. 

Busy for the rest of the day but tomorrow I should have some time to continue. I'll ping you on slack then 👍

test/__snapshots__/index.js.snap
```diff
@@ -67,6 +67,54 @@ const Test = () => <div className={\\"jsx-2529315885\\"}>
...
+const Test1 = () => [<div className={\\"jsx-336794318\\"}>Foo</div>, <span className={\\"jsx-336794318\\"}>Bar</span>, <_JSXStyle styleId={\\"1486255031\\"} css={\\"div.jsx-336794318{color:red}\\"} />];
+
+const Test2 = () => [<p className={\\"jsx-136928739\\"}>Hello World</p>, <_JSXStyle styleId={\\"2639504339\\"} css={\\"p{color:white}\\"} />];
+
+const Test3 = () => [<div className={\\"jsx-3360849707\\"}>
+    <p className={\\"jsx-3360849707\\"}>Hello World</p>
+    <_JSXStyle styleId={\\"4193872369\\"} css={\\"p.jsx-3360849707{color:blue}\\"} />
+  </div>, <div>Hello World</div>];
+
+const Test4 = () => [<div className={\\"jsx-612908104\\"}>
```
dsifford: Alright @giuseppeg so to start off I wanted to get a new baseline of where we currently stand.

After upgrading the dependencies and the snapshots, it looks like Test4 is the only place where it seems to still be stumbling. Specifically, the issue is that there is a decision between what style to scope into the selector if there are both a top-level style element and inner style elements.

I'll ping you on slack to discuss what you want to do about cases like this.

nexdrew:
Hi 👋 just wondering if there is any update on this, now that Next.js 4 is latest and uses React 16? Seems odd to specify different peer dependency versions between `next@latest` (currently 4.0.3) and `styled-jsx@latest` (currently 2.0.2).

Note that I'm not experiencing any problems, was just hoping to get rid of the npm/yarn warning. Thanks!

giuseppeg:
@nexdrew I've been busy at work, I'll see what I can do this weekend/next week

nexdrew:
@giuseppeg No worries, you've been doing some awesome work! I was hoping the update would be fairly straightforward, but I see there's more involved. If there's anything specific I can help with, just let me know. Thanks again!

amrdraz:
any work on this?

giuseppeg:
@dsifford I think that I am going to close this one. People now can use `React.Fragment` which in many cases is preferable to returning an array - I'd consider that a special case that we don't support. Thank you so much for your work and sorry if I couldn't help out.

Issue #247, opened 2017-06-26
Switch from data attributes to className


giuseppeg:
To improve performances we should switch to use `className` instead of the `data-jsx` attribute.

When we have spread props eg. `{...props}` we would need to move the `className` attribute to the end and rewrite it so it applies classes on cascade.

```jsx
<div {...props} className="foo" {...rest} />
// compiles to
<div {...props} {...rest} className={`${rest.className != null && rest.className || 'foo'} jsx123`} />
```

I actually started to work on it (didn't go far though so contribution are still welcome!) and here is the test case:

```jsx
export default () => (
  <div className='test'>
    <div className={test && test()} />
    <div className={test && 'test'} />
    <div className={`test ${test ? 'test' : ''}`} />
    <div className={a + 'test'} />
    <div className={`test ${test ? 'test' : ''}`} {...props} />
    <div className='hi' {...props} {...rest} />
    <div {...props} />
    <div {...props} {...rest} />
    <div />
    <style jsx>{'div { color: red }'}</style>
  </div>
)
```

Which would compile to (just showing some examples not all of them):

```jsx
// without className and spreads
<div className={`jsx123`} />

// with existing className
<div className={`test jsx123`} />

// with existing className and a spread
<div {...props} className={`${props.className != null && props.className || 'test'} jsx123`} />

// with existing className and many spread
<div {...props} {...rest} className={`${props.className  != null && props.className || rest.className != null && rest.className || `test ${test ? 'test' : ''}`} jsx123`} />

// with a spread only
<div {...props} className={`${props.className != null && props.className || ''} jsx123`} />

// with many spread only
<div {...props} {...rest} className={`${props.className != null && props.className || rest.className != null && rest.className || ''} jsx123`}  />
```

where `jsx123` is the style-jsx unique classname.

nkzawa:
SGTM

pruhstal:
@giuseppeg this still in progress?

giuseppeg:
@pruhstal I am working on #81 right now, this will probably be the next improvements

giuseppeg:
Fixed by #288

PR #288, opened 2017-09-27
styled-jsx v2

giuseppeg:
## Premise

My initial intention was to implement support for dynamic styles but along the way I had to implement a few related changes which made the scope of the PR grow a bit. This's why I think that this branch should mark the release of styled-jsx 2.0. I would also release a beta so we can get some people to try it first and fix bugs if there is any.

Probably it is nuts to try to review this PR, just read on, see if you like the API and take a look at the tests.

## What is this branch about

Mainly **introduce dynamic styles** support.

### switched from data attributes to className

While developing dynamic styles (see below) and benchmarking at some point I hit a wall and couldn't get any significant speed improvement.
After some profiling it was obvious that data-attributes made everything slower so I switched to `className`.

Now jsx is rewritten like so:

```jsx
export default () => (
  <div className="jsx-123" data-jsx>
    <_JSXStyle styleId="123" css={`div { color: red }`} />
  </div>
)
```

- [x] TBD - probably after transpilation we could remove the `data-jsx` attribute which is used to mark the path so that we don't process it twice by mistake.

See #247 and the snapshots from this PR for details.

### dynamic styles

Finally people will be able to use `props`, `state` and any other dynamic value in their styles like so:

```jsx
export default ({color, ...propsRest}) => (
  <div>
    <style jsx>{`
      div {
        display: inline-block;
        color: ${color};
        background-color: ${propsRest.backgroundColor}
      }
    `}</style>
  </div>
)
```

It works with any dynamic value (`this.props.color` included) and as opposed to other libs we don't need functions e.g. `color: ${props => props.color}` since we already have a component and we know which interpolations use props.

The old behavior is unchanged so interpolations that come from outside of the `render` method scope (or the function in case of a stateless fn component) are treated as constants. If any of the interpolations uses an identifier from the scope then styles are marked as dynamic.

The behavior as it was in v1 is unchanged therefore speed is still the same - actually it got better because we switched to `className` and introduced the `optimizeForSpeed` mode (see below).

Dynamic styles computation and classname computation is memoized which together with `optimizeForSpeed` makes it very fast ~60fps. Performance-wise this solution is slightly faster than emotion v7 and many other implementations.

<details>
  <summary>Implementation details</summary>

  When dynamic styles are detected (in the way described above) JSX is transpiled as follow:

  ```jsx
  export default ({color, ...propsRest}) => (
    <div className={_JSXStyle.dynamic([['123', [color, propsRest.backgroundColor]]])}>
      <JSXStyle
        styleId="123"
        css={`
          div.__jsx-dynamic-selector {
            display: inline-block;
            color: ${color};
            background-color: ${propsRest.backgroundColor}
          }
        `}
        dynamic={[color, propsRest.backgroundColor]}
      />
    </div>
  )
```

At runtime `_JSXStyle.dynamic` computes the class names using the base id `123` and the props array `[color, propsRest.backgroundColor]` which by then is just an array of strings.

`JSXStyle` does the same using the `styleId` and `dynamic` props. Once the classname is computed then it replace `__jsx-dynamic-selector` with the generated dynamic classname.

The class name computation as well as replaces are memoized to avoid repetitive work to happen.

With `optimizeForSpeed` which uses the CSSOM api and it is blazing fast (this is what glamor, emotion, glamorous use) `css` is an array of rules instead since `sheet.insertRule` accepts only one rule at the time.
</details>

### optimizeForSpeed

As mentioned before in order to make dynamic styles insertion fast we now use the CSSOM apis when the babel plugin `optimizeForSpeed` is set to `true` or if never set when `process.env.NODE_ENV === 'production'`. We only use this version in production since it has 2 cons:

* styles can't be edited via dev tools
* source maps don't work with this api

^ this is what libraries like emotion, glamor, glamorous and even styled-components (iirc) do.

Obviously in dev you get all the above and things are still quite fast.

<details>
  <summary>Implementation details</summary>
  In order to achive this I got rid of the old `render.js` and refactored `style.js`.

  I created a standalone `StyleSheet` class in `src/lib/stylesheet.js` which abstracts away the differences between CSSOM api and normal style tags in addition to support SSR.

  Then I created a `StyleSheetRegistry` which takes care of deduping / keeping instances around and inserting/deleting/replacing css rules.

  Finally for the CSSOM `sheet.insertRule` api to work we need to split the CSS blocks into single rules e.g.

  ```js
  `div { color: red } span { color: hotpink }`
  ```

  becomes an array of rules:

  ```js
  ['div { color: red }', 'span { color: hotpink }']
  ```

  This is done with a Stylis plugin in `src/lib/styles-transform.js`
</details>

### External styles now must be tagged with css``

Previously we did some black magic to try to detect exported styles. Basically every exported string was parsed by CSSTree and if it passed validation was marked as "style". The main cons of this is that CSSTree would fail silently when the CSS was invalid and use couldn't tell why their styles were ignored.

To fix this now template literals need to be tagged with `css`.

```js

export default css`
  div { color: red }
`
```

As a consequence of that people using Prettier 1.6.0 will also get support for external styles automatically!

### TODO
- [x] Fix typos


giuseppeg:
> TBD - probably after transpilation we could remove the `data-jsx` attribute which is used to mark the path so that we don't process it twice by mistake.

I decided to remove the data attribute. Not sure if it will ever be the case that a path is visited twice!?

rstacruz:
This is amazing. Love the work all you guys are doing for styled-jsx! It'd be nice to have a benchmark soon.

mxstbr:
> and even styled-components (iirc) do.

Nah, that's why we're so "slow" in microbenchmarks. 😜 Since lots of people asked for it we're going to implement it sometime soon though.

hanford:
Very exciting PR! 

I switched from `styled-jsx` to `styled-components` for this exact reason, but my company plans on transitioning to NextJS and the `styled-jsx` integration is too good to not use.

readme.md
```diff
@@ -182,9 +214,56 @@ export default () => (
...
+          display: inline-block;
+          font-size: 1em;
+        }
+     `}</style>
+  </button>
+)
+```
+
+New styles injection is optimized to perform well at runtime and maintain 60fps.
+Nonetheless we recommend to split static and dynamic styles when possible. The following will be even faster:
```
nkzawa: Is this because we need some dynamic computations ? It might be nice if there is a explanation a bit.

giuseppeg: If most of the CSS is static it would be ideal to use two `style` tags so that only the dynamic parts are recomputed/rendered. In fact when in dev mode we could print some message / warning when we detect similar situations (maybe in a followup)

PR #292, opened 2017-10-03
Fix hashes should match when using a single tag

giuseppeg:
When there is **only a single** `style` tag per component we still `hashString` its hash (since we always do `hashString(hashes)` where `hashes` is an array). The resulting hash is used to scope CSS. 

However the `styleId` for `JSXStyle` then used just `hashes[0]` resulting in something like this:

```jsx
<_JSXStyle styleId="1234" css="div.jsx-9887 { color: red }" />
```

Instead with this fix it will match:

```jsx
<_JSXStyle styleId="9887" css="div.jsx-9887 { color: red }" />
```

Again, this was an issue only when there was a single `style` tag per component. When there are multiple it is normal that `styleId` and the hash in the css don't match

### DIFF
--- a/src/_utils.js
+++ b/src/_utils.js
@@ -19,6 +19,8 @@ const or = (a, b) => t.logicalExpression('||', a, b)
 
 const joinSpreads = spreads => spreads.reduce((acc, curr) => or(acc, curr))
 
+export const hashString = str => String(_hashString(str))
+
 export const addClassName = (path, jsxId) => {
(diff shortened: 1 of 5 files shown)


giuseppeg:
cc @timneutkens

timneutkens:
Could you do a new release after merging. Then we can add it to next v4 👌

giuseppeg:
@rauchg I should have tested with next-news before merging. Did you find any bug? I just saw https://github.com/zeit/styled-jsx/tree/revert-292-fix-hash

PR #304, opened 2017-10-10
Fix tagged template literals detection

### LINKED ISSUE #303: Version 2 seems to be conflict with styled-component
v1 works perfectly with the `.extend` in styled-component,it seems that style-jsx transform the styled-component template Literals

more details in https://github.com/styled-components/styled-components/issues/1228

giuseppeg:
Fixes #303

Tagged template literals detection for external files was buggy and unfortunately transpiling arbitrary 3rd parties strings like styled-component's ones.

While I was there I rewrote it to work like babel-macros which should make it more reliable.


src/babel-external.js
```diff
@@ -128,28 +113,37 @@ function makeHashesAndScopedCssPaths(exportIdentifier, data) {
...
-        sourceFileName: state.file.opts.sourceFileName,
-        sourceMaps: state.file.opts.sourceMaps
-      },
-      splitRules:
-        typeof state.opts.optimizeForSpeed === 'boolean'
-          ? state.opts.optimizeForSpeed
-          : process.env.NODE_ENV === 'production'
+    const tagName = path.node.specifiers[0].local.name
+    const taggedTemplateExpressions = path.scope
+      .getBinding(tagName)
```
giuseppeg: need to check that this is not undefined

PR #313, opened 2017-10-15
Use stylis-rule-sheet plugin

giuseppeg:
Replace our plugin to split rules with stylis-rule-sheet


giuseppeg:
@thysultan do you mind reviewing this branch?

src/lib/style-transform.js
```diff
@@ -152,16 +97,15 @@ function transform(hash, styles, settings = {}) {
...
-  const cssString = stylis(scopeHash, styles)
+  stylis(hash, styles)
 
-  if (isSplitRulesEnabled) {
+  if (settings.splitRules) {
     return splitRules
   }
-  return cssString
+
+  return splitRules.join('')
```
giuseppeg: @thysultan as you can see from the diff before I was returning an array of `splitRules` only when `settings.splitRules` was set to `true` otherwise we'd just return the processed string. With stylis-rule-sheet the string contains `/*|*/` therefore I decided to return `splitRules.join('')` - do you see any cons of doing so?

PR #316, opened 2017-10-16
Add less plugin link

erasmo-marin:
I have been working on a less plugin as a proof of concept, I think it would be added to readme.

As less compiler is async, I reused some code from postcss plugin.


giuseppeg:
Awesome! thank you @erasmo-marin

Issue #319, opened 2017-10-17
StyleSheetRegistry errors


slaskis:
I believe we found a regression in 2.0 which only shows up in production.

It seems to be some kind of cleanup done by styled-jsx 2.0. Where empty ```<style jsx>{``}</style>```-tags are used sometimes but removed when navigating away which causes this error:

```
Error: StyleSheetRegistry: styleId: `jsx-2085888330` not found.
```

I've added [a commit](https://github.com/slaskis/styled-jsx-test/commit/69e2e6b595bf70f96f9019951153d618f6c664f2) to my sample repo which shows the issue when you navigate back and forth between the "pages".

pramjeet:
I am also facing this issue in production, does anybody know a quick fix for now please?

giuseppeg:
The quick fix is to not use the `optimzeForSpeed` mode (or rather set it to `false`).

@slaskis why would you use an empty style tag?

pramjeet:
I found some empty `<style jsx>{``}</style>` and just removed them. It fixed the problem for me.

@giuseppeg Thanks anyway.

slaskis:
I just updated next to 4 and this started to happen in production.

I don't know why but a colleague of mine does it kind of like a template for new components which will get styles later.

I would understand if it's not supported but a better and maybe more consistent error would be nice. Now it's a pretty aggressive regression which occur when you upgrade.

giuseppeg:
yup we could filter out empty rules of course :) PR are welcome!

rockchalkwushock:
FYI beginning work on this, wish me luck 😂  🎃 💩

williamkunz:
@slaskis Could this be your issue? https://github.com/zeit/styled-jsx/pull/327#issuecomment-338312692

slaskis:
Looks very interesting!

bathos:
I’m seeing this error on navigation in IE11 sometimes even though I have no empty Styled JSX elements.

giuseppeg:
As I said to @williamkunz 

> yeah whenever there is a parse error the rule is ignored. This happens only when optimizeForSpeed is enabled (usually automatically for production builds). To prevent surprises in prod you may want to enable optimizeForSpeed manually (via babel plugin options) while you are developing. When optimizeForSpeed is enabled and process.env.NODE_ENV !== 'production' you should get a warning.

Ideas to improve this are welcome.

bathos:
[...]

PR #327, opened 2017-10-20
filter out undefined rules

williamkunz:
I came across this issue during a Friday (today) prod push, of course. This was a 2 minute fix but has seemingly fixed issue #319 

I haven't deep dived into the source so if there is a more elegant fix required, feel free to close.

### DIFF
--- a/src/stylesheet-registry.js
+++ b/src/stylesheet-registry.js
@@ -64,6 +64,11 @@ export default class StyleSheetRegistry {
 
   remove(props) {
     const { styleId } = this.getIdAndRules(props)
+
+    if (!this._instancesCounts[styleId]) {
+      return false
+    }
+
     invariant(
       styleId in this._instancesCounts,
       `styleId: \`${styleId}\` not found`


williamkunz:
I'll keep digging to see if there is a better solution

williamkunz:
Upon further digging, 319 may be a non-issue.

From converting Glamor code we had an external string:
```js

export const myStyles = css`
  .my-class: {
    padding: 1.6rem;
  }
`
```
vs. the correct:
```js

export const myStyles = css`
  .my-class {
    padding: 1.6rem;
  }
`
```

The extra colon was just leftover from the conversion and causing the same error as described in #319. Maybe this issue should become an optimization with some warning within the `css` module?

giuseppeg:
@williamkunz yeah whenever there is a parse error the rule is ignored. This happens only when `optimizeForSpeed` is enabled (usually automatically for production builds). To prevent surprises in prod you may want to enable `optimizeForSpeed` manually (via babel plugin options) while you are developing. When `optimizeForSpeed` is enabled and `process.env.NODE_ENV !== 'production'` you should get a warning.

Issue #329, opened 2017-10-20
Rewrite complex spread props breaks


ndreckshage:
Wasn't 100% sure how to title this, but...

compile error `Cannot read property '0' of null`
```
const Component = props => {
  return (
    <div>
      <div
        {...(props.something
          ? {}
          : {
              onClick: () => {
                alert(1);
              }
            })}
      >
        hello
      </div>
      <style jsx>{`
        div {
          color: blue;
        }
      `}</style>
    </div>
  );
};
```

no compile error:
```
const Component = props => {
  const sameThingAsVar = props.something
    ? {}
    : {
        onClick: () => {
          alert(1);
        }
      };

  return (
    <div>
      <div {...sameThingAsVar}>hello</div>
      <style jsx>{`
        div {
          color: blue;
        }
      `}</style>
    </div>
  );
};
```

ndreckshage:
btw in my apps im just pulling up into a var. but wouldnt be surprised if others porting a reasonable big project from v1 to 2. but other than that seems to be a seamless upgrade - amazing!

giuseppeg:
@ndreckshage thanks for the bug report. The issue is indeed with such a complex spread expression. 
I will try to make it work for complex cases like that too. In the meantime your solution is what I would I have gone for too

guigrpa:
I have a simpler repro. Indeed there seems to be a problem with the spread operator, but also in simpler cases.

**This compiles:**

```js
import React from 'react';
const A = props => (
  <div className="a-component" {...props}>
    <style jsx>{`.a-component { color: red; }`}</style>
  </div>
);
export default A;
```

**But this doesn't:**

```js
import React from 'react';
class A extends React.Component {
  render() {
    return (
      <div className="a-component" {...this.props}>
        <style jsx>{`.a-component { color: red; }`}</style>
      </div>
    );
  }
}
export default A;
```

**A workaround that compiles again:**

```js
import React from 'react';
class A extends React.Component {
  render() {
    const { props } = this;
    return (
      <div className="a-component" {...props}>
        <style jsx>{`.a-component { color: red; }`}</style>
      </div>
    );
  }
}
export default A;
```

giuseppeg:
good point @guigrpa @ndreckshage would you have time to at least fix the `{...this.props}` case? It shouldn't be hard.. here is the code which alterate the attributes https://github.com/zeit/styled-jsx/blob/master/src/_utils.js#L24-L78 

[...]

PR #356, opened 2017-11-27
Fix complex spread props breaking (#329)

pudility:
Fixing issue #329. 

TODO:

- [X] Add test cases for `this.props.foo`
- [X] Unlimited length `MemberExpression`s


src/_utils.js
```diff
@@ -31,10 +31,18 @@ export const addClassName = (path, jsxId) => {
     const node = attr.node
     if (t.isJSXSpreadAttribute(attr)) {
       const name = node.argument.name
-      const attrNameDotClassName = t.memberExpression(
+      let attrNameDotClassName = t.memberExpression(
         t.identifier(name),
         t.identifier('className')
       )
+
+      if (node.argument.type === 'MemberExpression') {
```
giuseppeg: here you can do the following:

```js
let attrNameDotClassName

if (t.isMemberExpression(node.argument)) {
   attrNameDotClassName = // with nested member expression
} else {
   attrNameDotClassName = // with simple identifier
}
```

test/__snapshots__/index.js.snap
```diff
@@ -81,6 +81,47 @@ export default class {
 }"
 `;
 
+exports[`works with complex spread props (class) 1`] = `
```
giuseppeg: Can you revert the changes to this file? It seems that you removed the fixture and the test from `test/index.js` but the snapshot is still here in the diff :)

pudility: I added back the tests in `test/index.js`, is that what you where looking for? If not I am not sure I understand what you are asking, sorry.

test/index.js
```diff
@@ -20,6 +20,13 @@ test('works with stateless', async t => {
   t.snapshot(code)
 })
 
+test('works with spread', async t => {
```
giuseppeg: we don't need this test here since it is already in `test/attribute.js`

test/__snapshots__/index.js.snap
giuseppeg: Sorry I didn't remember correctly the tests for this feature are in `test/attribute.js`. 
Can you do the following:

* remove the test from `test/index.js`
* regenerate the snapshot for `test/index.js` like so:

```
rm test/__snapshots__/index.js.snap
npm test -- -u
```

src/_utils.js
```diff
@@ -31,10 +31,20 @@ export const addClassName = (path, jsxId) => {
...
+        attrNameDotClassName = t.memberExpression(
+          node.argument,
+          t.identifier('className')
+        )
+      } else {
+        attrNameDotClassName = t.memberExpression(
+          t.identifier(name),
+          t.identifier('className')
+        )
+      }
```
giuseppeg: nitpick: I just realized that this could be:

```js
const attrNameDotClassName = t.memberExpression(
  t.isMemberExpression(node.argument) ? node.argument : t.identifier(name),
  t.identifier('className')
)
```

pudility:
I think I got it all, but if there is anything else I can change, just let me know :)

giuseppeg:
@pudility awesome, thank you again for contributing!

pudility:
@giuseppeg No problem, thanks for all the mentoring.

PR #362, opened 2017-12-08
Update to Babel 7.x

benedyktdryl:
Hi, 

This is my very firs PR which includes Babel 7.x related updates. Please let me know if you will be interested in such contribution so I can make further updates in that matter.

### DIFF
--- a/benchmark/babel.js
+++ b/benchmark/babel.js
@@ -1,7 +1,7 @@
 import {readFileSync} from 'fs'
 import {resolve} from 'path'
 import {Suite} from 'benchmark'
-import {transform as babel} from 'babel-core'
+import {transform as babel} from '@babel/core'
 
 import plugin from '../src/babel'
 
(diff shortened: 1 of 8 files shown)


giuseppeg:
@benedyktdryl thanks! why not..

benedyktdryl:
Thanks. Question: is there any chance this can be released under some beta tag? If yes, it will be easier to maintain changes in Next.js related PR:

https://github.com/zeit/next.js/pull/3428

giuseppeg:
Oh sorry I forgot we had babel related `dependencies`. That is going to be an issue for the ones who don't use Babel 7 yet right? If that's the case maybe we can wait a bit or if Next.js needs styled-jsx to be on Babel 7 then we could release a separate, temporary version. Maybe dist-tag it with babel7

benedyktdryl:
Indeed there will be an issue, an dist-tag is good idea, thanks.

giuseppeg:
will reconsider this when Babel 6 is deprecated or we find a better way to guarantee backwards compatibility

Issue #366, opened 2017-12-13
Throws `Unfortunately nesting is not supported by styled-jsx` when using CSS comments before @media queries


antonj:
Added failing test here https://github.com/antonj/styled-jsx/commit/c357234a331dc971f88ca667a8291822a18c703b

```
div { color: red; }
/* CSS Comment */
@media screen and (min-width: 480px) {
  div { color: red; }
}
```

Throws `Unfortunately nesting is not supported by styled-jsx` if `/* CSS Comment */` is removed the tests pass.

Pro542:
Tried to include a local styled-jsx plugin but even `(css, options) => css` which does not transform the css gives the nesting error

jabacchetta:
Has anyone figured out a fix for this yet? I just did a fresh `npm i` and now I can't get back to a working state at all.

a-ignatov-parc:
I'm using `styled-jsx` with `stylus` plugin and there are no problems with comments in styles. Maybe this could be a workaround until it gets fixed.

P.S. If you don't want to use `stylus` or other preprocessors you can try to use [postcss-discard-comments](https://www.npmjs.com/package/postcss-discard-comments) with [postcss plugin](https://github.com/giuseppeg/styled-jsx-plugin-postcss) just to discard comments ;)

giuseppeg:
will look into this in the following days

jabacchetta:
@giuseppeg Just FYI, I had a few minutes to play around with this today. I don't believe it has anything to do with comments.

I copy/pasted what @antonj had without the comments and it did work. But then I attempted to modify my own media query (which did not have any comments but was failing), and I had it down to exactly what he had, yet it still failed. Then I started copy pasting in bits from his code and messing with spacing and eventually got it to work.

So it seems as though spacing has something to do with it. I'll try to experiment more tonight if I have the time.

jabacchetta:
@giuseppeg OK, for me, I simply had to remove the spacing that was on the line break immediately preceding the media query declaration. That fixed it, although no idea why yet. Hopefully that helps somewhat.

cpugirl:
Yes, it worked . Just remove extra spaces before media query.!!

giuseppeg:
@antonj I tried to add that css to the tests but they still pass. Do you mind creating a pull request with your code? Do you use tabs for indentation?

@jabacchetta same with your code.

antonj:
[...]

PR #372, opened 2017-12-22
Add failing test: throws when using a line with white-space (tab or space) before at-rules

antonj:
Related to #366


giuseppeg:
thanks

PR #375, opened 2017-12-27
Properly expose filename to plugins

MarcoThePoro:
I'm trying to add relative imports to the styled jsx plugin `styled-jsx-plugin-sass`, but to do that, I need to be able to give sass the directory containing the file (that contains the CSS chunk).

You can read more of the discussion here: https://github.com/giuseppeg/styled-jsx-plugin-sass/issues/10

I imagine other plugins could have their uses for this too, if anything is wrong feedback is appreciated!

I noticed when I was testing this via nextjs that `fileInfo.sourceFileName` can have a query attached to it eg `./pages/index.js?entry` and I wasn't sure whether to strip the query or leave that to the plugin author, in this PR I just left it.

Thanks!

(EDIT: oh no snapshot testing I don't know how to deal with it!)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -553,7 +553,6 @@ The `options` object has the following shape:
     vendorPrefixes: boolean,
     isGlobal: boolean,
     filename: ?string, // defined only when styled-jsx/babel is used via Babel CLI,
-    sourceFileName: string, // the original Javascript file the CSS block was in
     location: { // the original location of the CSS block in the JavaScript file
       start: {
         line: number,
--- a/readme.md
+++ b/readme.md
@@ -553,6 +553,7 @@ The `options` object has the following shape:
     vendorPrefixes: boolean,
     isGlobal: boolean,
     filename: ?string, // defined only when styled-jsx/babel is used via Babel CLI,
+    sourceFileName: string, // the original Javascript file the CSS block was in
     location: { // the original location of the CSS block in the JavaScript file
       start: {
         line: number,
--- a/readme.md
+++ b/readme.md
@@ -552,7 +552,7 @@ The `options` object has the following shape:
     sourceMaps: boolean,
     vendorPrefixes: boolean,
     isGlobal: boolean,
-    filename: ?string, // defined only when styled-jsx/babel is used via Babel CLI,
+    filename: ?string, // defined only when the filename option is passed to Babel, such as when using Babel CLI or Webpack
     location: { // the original location of the CSS block in the JavaScript file
       start: {
         line: number,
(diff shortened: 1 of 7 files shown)


src/_utils.js
```diff
@@ -524,7 +524,8 @@ export const processCss = (stylesInfo, options) => {
     vendorPrefixes,
     sourceMaps: useSourceMaps,
     isGlobal,
-    filename: fileInfo.filename
+    filename: fileInfo.filename,
```
giuseppeg: Mh can't remember what is the difference between `filename` and `sourceFileName` 🤔 Do you mind checking in the babel docs or trying to console.log both in a plugin? Maybe `filename` is enough?

readme.md
```diff
@@ -553,6 +553,7 @@ The `options` object has the following shape:
     vendorPrefixes: boolean,
     isGlobal: boolean,
     filename: ?string, // defined only when styled-jsx/babel is used via Babel CLI,
+    sourceFileName: string, // the original Javascript file the CSS block was in
```
giuseppeg: this is also defined _when styled-jsx/babel is used via Babel CLI_.
i.e. if you use `transform` from `babel-core` and pass it the file content it won't have any of the `file` information.

giuseppeg: basically you need to prefix `string` with a `?` and copy the comment above.

MarcoThePoro: Hmm when I was working on this I was testing it by editing the package in my node_modules of a nextjs project, fileInfo.sourceFileName seemed to be present but fileInfo.filename wasn't, I'll check the Babel docs to be sure

MarcoThePoro:
So I did some investigating and here's what I found.

- As far as I can tell, `state.file.filename` is never defined, and what we're really after is `state.file.opts.filename`. (`state.file.filename` is placed onto fileInfo [here](https://github.com/zeit/styled-jsx/blob/940332a5bd3b3b91bfc4162cd72aff7c53e543e0/src/babel.js#L241))
- `sourceFileName` is the relative path from the file to the source map, I shouldn't have been using it! (it just happened to be the same path I was looking for when compiling via webpack)
- According to [the babel docs](https://babeljs.io/docs/usage/api/#options) `state.file.opts.filename` comes from the options passed to babel, so it's not guaranteed to be present. It is present when compiling via babel-cli and webpack, as well as any other sensible tool using babel I imagine.

I found this by logging some stuff in a babel plugin as you suggested, you can have a look: https://github.com/MarcoThePoro/babel-fileinfo-test

giuseppeg:
@MarcoThePoro sorry I've been busy (or rather afk) I'll see if I can take a look at this pr over the weekend.

PR #387, opened 2018-01-16
Update info about autocomplete for vscode

Grimones:
Hi there! First of all thanks for the `styled-jsx`! it is really awesome solution which solved some nontrivial things for me. 

Now it has autocomplete in vscode 🎉 

I have made a language server for `styled-jsx` which works the same as stock vscode `CSS` language server (they are identical 😅)

I will be really happy if you will merge changes for readme where I mention my extension.
Thank you!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -706,13 +706,17 @@ const Button = ({ children }) => (
 }
 ```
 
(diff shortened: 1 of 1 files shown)


giuseppeg:
@Grimones awesome thank you!!

PR #397, opened 2018-01-30
Fixed dynamic styles detection

a-ignatov-parc:
Fixed case when dynamic styles with compound props was detected as static

### DIFF
--- a/src/_utils.js
+++ b/src/_utils.js
@@ -129,11 +129,13 @@ export const validateExpressionVisitor = {
     }
   },
   Identifier(path, scope) {
-    if (t.isMemberExpression(path.parentPath)) {
+    const { name } = path.node
+
+    if (t.isMemberExpression(path.parentPath) && scope.hasOwnBinding(name)) {
       return
     }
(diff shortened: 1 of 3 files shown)


a-ignatov-parc:
@giuseppeg is there any chance we can get this fix released till Friday? 

Don't want to rollback migration to `styled-jsx` just before release :)

giuseppeg:
@a-ignatov-parc hi sorry, looks good. 

Can you add two more tests:

```jsx
const foo = 'red'
() => <div><style jsx>{`div { color: ${foo} }`}</style></div>
```

and one like the one you added but without the `if`? (without a block)

a-ignatov-parc:
Sure 👍

a-ignatov-parc:
Done. Looks good.

giuseppeg:
thanks man, I'll publish a patch release right away

a-ignatov-parc:
Awesome! Thanks a lot 🎉

giuseppeg:
done

PR #403, opened 2018-02-06
Sanitize dynamic styles

giuseppeg:
We are not going to sanitize dynamic props on the client when using CSSOM API since we don't inject dom/strings.

### DIFF
--- a/src/stylesheet-registry.js
+++ b/src/stylesheet-registry.js
@@ -153,10 +153,10 @@ export default class StyleSheetRegistry {
   ) {
     const cache = {}
     return function(id, css) {
-      // Sanitize SSR-ed and client side (style tags) CSS.
-      // When using optimize for speed we don't need to sanitize
-      // because we are not inserting markup in the DOM.
-      if (!this._isBrowser || !this._optimizeForSpeed) {
+      // Sanitize SSR-ed CSS.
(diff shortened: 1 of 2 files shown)


src/stylesheet-registry.js
```diff
@@ -146,6 +153,12 @@ export default class StyleSheetRegistry {
   ) {
     const cache = {}
     return function(id, css) {
+      // Sanitize SSR-ed and client side (style tags) CSS.
+      // When using optimize for speed we don't need to sanitize
+      // because we are not inserting markup in the DOM.
+      if (!this._isBrowser || !this._optimizeForSpeed) {
```
timneutkens: Why not in the browser? Just curious 🙏

giuseppeg: !this._isBrowser == server side rendering

giuseppeg: indeed I can also remove the second `!this._optimizeForSpeed`. Client side code should be safe because of `document.createTextNode` (dev) and the CSS OM api not injecting DOM (prod)

giuseppeg:
@timneutkens just released `2.2.4` which includes this fix.

timneutkens:
@giuseppeg will upgrade in next 🙏

PR #421, opened 2018-03-17
Modernize things

giuseppeg:
I am removing gulp, babel-polyfill which we only needed because we used a `Map` in `src/style.js`.
I replaced the Map with an array. Also switched to babel-preset-env. Finally I am removing the benchmarks since there are better ones around http://necolas.github.io/react-native-web/benchmarks/ https://github.com/A-gambit/CSS-IN-JS-Benchmarks


src/style.js
```diff
@@ -6,10 +6,9 @@ const styleSheetRegistry = new StyleSheetRegistry()
 export default class JSXStyle extends Component {
   static dynamic(info) {
     return info
-      .map(tagInfo => {
-        const [baseId, props] = tagInfo
```
giuseppeg: We can just get rid of this and Babel won't inline the following helper :)

```js
var _slicedToArray = (function() {
  function sliceIterator(arr, i) {
    var _arr = [];
    var _n = true;
    var _d = false;
    var _e = undefined;
    try {
      for (
        var _i = arr[Symbol.iterator](), _s;
        !(_n = (_s = _i.next()).done);
        _n = true
      ) {
        _arr.push(_s.value);
        if (i && _arr.length === i) break;
      }
    } catch (err) {
      _d = true;
      _e = err;
    } finally {
      try {
        if (!_n && _i["return"]) _i["return"]();
      } finally {
        if (_d) throw _e;
      }
    }
    return _arr;
  }
  return function(arr, i) {
    if (Array.isArray(arr)) {
      return arr;
    } else if (Symbol.iterator in Object(arr)) {
      return sliceIterator(arr, i);
    } else {
      throw new TypeError(
        "Invalid attempt to destructure non-iterable instance"
      );
    }
  };
})();
```

src/style.js
```diff
@@ -7,7 +7,8 @@ export default class JSXStyle extends Component {
   static dynamic(info) {
     return info
       .map(tagInfo => {
-        const [baseId, props] = tagInfo
```
giuseppeg: by getting rid of this Babel won't inline the following helper:
```js
var _slicedToArray = (function() {
  function sliceIterator(arr, i) {
    var _arr = [];
    var _n = true;
    var _d = false;
    var _e = undefined;
    try {
      for (
        var _i = arr[Symbol.iterator](), _s;
        !(_n = (_s = _i.next()).done);
        _n = true
      ) {
        _arr.push(_s.value);
        if (i && _arr.length === i) break;
      }
    } catch (err) {
      _d = true;
      _e = err;
    } finally {
      try {
        if (!_n && _i["return"]) _i["return"]();
      } finally {
        if (_d) throw _e;
      }
    }
    return _arr;
  }
  return function(arr, i) {
    if (Array.isArray(arr)) {
      return arr;
    } else if (Symbol.iterator in Object(arr)) {
      return sliceIterator(arr, i);
    } else {
      throw new TypeError(
        "Invalid attempt to destructure non-iterable instance"
      );
    }
  };
})();
```

hzoo: If you mean you want smaller output, you'll need to use other options like 'loose' mode because spec is default.

giuseppeg: good point but I don't know if I can configure that per-plugin when using preset-env. In this case I can just write es5 myself

hzoo: You can apply loose mode to all plugins or do them individually yeah. Of course that works too

PR #422, opened 2018-03-20
Rewrite styled-jsx/css

giuseppeg:
Since when we introduced it in v1 tagging a string with `css` would generate both scoped and global styles because there was no way for us to guess what were the user intentions.  This sucks because you'd get twice as much code in your bundles (ok gzip probably made this suck less but still it is not optimal).

I rewrote the entire external and outside of component styles  plugin to finally generate only scoped or global styles based on which tag. Basically now `css` by default generates scoped styles and global styles can be created using `css.global`:

```js

// Scoped styles
export const button = css`button { color: hotpink; }`

// Global styles
export const body = css.global`body { margin: 0; }`

// Works also with default exports
export default css`div { color: green; }`
```
In addition to avoid reduce the bundle size this change allows for one awesome thing that is... we could write a webpack loader or use [an existing one](https://www.npmjs.com/package/styled-jsx-css-loader) (I can talk to this guy) to allow people to write CSS in external `.css` files which means that they get all the tools they are used and can do all the kind of transformations they want before styled-jsx runs on them.

This branch introduces another neat feature that is a style resolver: `css.resolve` which yields an object that contains the scoped `className` and the scoped `styles` (a react element):

```jsx

// Resolved styles
export const link = css.resolve`a { color: green; }`
// link.className -> scoped className to apply to `a` elements e.g. jsx-123
// link.styles -> styles element to render inside of your component
```

This solution was basically [already possible in user land](https://github.com/zeit/styled-jsx/tree/2831798e31646133296feaa9096f66a15f4672bc#styling-third-parties--child-components-from-the-parent) but now we have a proper API for it, and it can also work with dynamic styles:

```js
function getStyles(color) {
  return css.resolve`
    a { color: ${color} }
  `
}
```

Final **bonus**. I create a [Babel macro](https://github.com/kentcdodds/babel-plugin-macros) for `resolve` which should allow people to use this last feature with Create React App.

Since these are breaking changes, I'd say we release v3.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -8,7 +8,7 @@
 Full, scoped and component-friendly CSS support for JSX (rendered on the server or the client).
 
 
-Code and docs are for v2 which we highly recommend you to try. Looking for styled-jsx v1? Switch to the [v1 branch](https://github.com/zeit/styled-jsx/tree/v1).
+Code and docs are for v3 which we highly recommend you to try. Looking for styled-jsx v2? Switch to the [v2 branch](https://github.com/zeit/styled-jsx/tree/v2).
 
 For an overview about the **features** and **tradeoffs** of styled-jsx you may want to take a look at [this presentation](https://speakerdeck.com/giuseppe/styled-jsx).
 
@@ -21,8 +21,6 @@ For an overview about the **features** and **tradeoffs** of styled-jsx you may w
 - [How It Works](#how-it-works)
   * [Why It Works Like This](#why-it-works-like-this)
 - [Targeting The Root](#targeting-the-root)
-- [Keeping CSS in separate files](#keeping-css-in-separate-files)
-- [Styles outside of components](#styles-outside-of-components)
 - [Global styles](#global-styles)
   * [One-off global selectors](#one-off-global-selectors)
 - [Dynamic styles](#dynamic-styles)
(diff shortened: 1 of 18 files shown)


jadengore:
@giuseppeg Question about this PR. Does `css.resolve` still have the same behavior where nested selectors are prefixed?

Use case question I have is when passing a `className` to a third-party component, and I want only the top-level selector prefixed and everything within it not prefixed.

giuseppeg:
@jadengore yes every selector in the css string is scoped. 

> Use case question I have is when passing a className to a third-party component, and I want only the top-level selector prefixed and everything within it not prefixed.

For this you don't need styled-jsx I guess. You can just generate a unique classname that you would use like in ID to style the third party component with global styles

timneutkens:
> we could write a Webpack loader or use an existing one (I can talk to this guy) to let people write CSS in external .css files.

cc @hanford is currently using https://github.com/traveloka/styled-modules

timneutkens:
Awesome!

jadengore:
When will this be released?

giuseppeg:
we are kinda blocked by https://github.com/thysultan/stylis.js/issues/101

readme.md
```diff
@@ -444,6 +386,161 @@ It's **paramount** that you use one of these two functions so that
...
+  <div>
+    <button>styled-jsx</button>
+    <style jsx>{styles}</style>
+    <style jsx>{button}</style>
+    <style jsx global>{body}</style>
+  </div>
+)
+```
+
+N.B. All the tags except for [`resolve`](#the-resolve-tag) don't support dynamic styles.
```
jkjustjoshing: What does "N.B." mean here?

PR #427, opened 2018-03-30
Placeholders tokens for plugins

giuseppeg:
This branch is an attempt to make the life of plugin authors' easier. 

The main pain point for plugins authors are expressions since those can be problematic when transforming the css using other parsers.

By detecting the expressions' types we can make a simple helper function that plugin authors can use to replace placeholders with valid CSS so that SASS/PostCSS/anything else works properly. Such a function would then return another function to restore the placeholders that the plugin author has to use before returning CSS from the plugin.

```js
import { replaceExpressions } from 'styled-jsx/utils'

function sassPlugin(css, settings) {
  const result = replaceExpressions(css, {
    SELECTOR: id => `._styled-jsx-${id}`,
    DECLARATION: id => `content: 'styled-jsx-placeholder-${id}'`,
    // ...     
  })  
  const sassCompliantCSS = result.css
  const compiled = sass.renderSync({ data: sassCompliantCSS }).css.toString()

  return result.restore(compiled)
}
```


giuseppeg:
@thysultan I would love to get a review from you on this, probably some day something like this could be built-in into Stylis. I still need to add more test cases.

thysultan:
Is valid CSS the only problem with placeholder expressions? Can you point me to an example that a parser might error on? Could you use persistent license comments `/*!..*/` as a universal no-op that parses will ignore but not discard?

giuseppeg:
technically we could even take care of replacing and restoring the placeholders with some valid css

giuseppeg:
@thysultan not sure if that would work e.g.

```css
div { color: /*!foo*/; }
```
has no value

thysultan:
Yes [sass](https://sass.js.org/) seems to ignore those. The following works on both sass and postcss:

```css
div {
  color:attr(--styled-jsx-id--);
  /* or */
  color:--styled-jsx-id--;
}
```
If you are able to target selectors that might look like

```css
[--styled-jsx-id--] {
  color: red;
}
```

Though linter plugins might not play well with that value type. Is there any point in the lifecycle of the pre-processing context that styled-jsx is holding onto raw valid css?

giuseppeg:
As soon as there are expressions there won't be valid CSS (at compile time).

thysultan:
Since following is symmetrically valid css if you know when an expression is a selector you could use the following as a placeholder. Not sure what other linters apart from stylelint would make of this.

```css
div {
  color: \\2D;
  /* or */
  color: /* stylelint-disable */\\2D/* stylelint-enable */;
  /* but even stylelint seems to accept just \\2D */
}

[\\2D] {
  color: red;
}
```

giuseppeg:
yeah finding good enough placeholder is not an issue I guess. This PR is about determine what are the expressions for and assigning them a `TYPE` - not a trivial task

thysultan:
Stylis plugins receive contextual info on what type(selector/value/etc) of expression a particular part is, is there any reason this could not be archived with a stylis plugin?

giuseppeg:
probably it'd be easier to pass the css with placeholder to a custom PostCSS plugin, since PostCSS seems to be parsing mostly everything.

giuseppeg:
sorry async commenting :D So stylis can parse our CSS with placeholder and be able to tell what is the placeholder? eg:

```css
div {
  color: %%styled-jsx-placeholder-1%%;
}
```
Can it tell that `%%styled-jsx-placeholder-1%%` is a `value`? Obviously this is the simplest case.

giuseppeg:
Oh I forgot that at this stage the CSS we get can be non canonical, so even my attempt to detect the expressions type can be flaky ... very easily. Eg. when we get Stylus code. Still PostCSS seems to be parsing everything I feed to it :o 

We would use Stylis later after we've passed all the user plugins, to scope our classes and add prefixes

thysultan:
Yes.

```scss
%%styled-jsx-placeholder-1%%,
%%styled-jsx-placeholder-2%% {
  color: %%styled-jsx-placeholder-3%%;
}
```

It can tell that 1 and 2 are selectors and 3 is a property value.

giuseppeg:
Going to close this. Filed an issue https://github.com/zeit/styled-jsx/issues/431

PR #438, opened 2018-05-02
chore: ignore .DS_Store files

DanielRuf:


giuseppeg:
Thanks for the PR! For this usually you can use a global gitignore https://help.github.com/articles/ignoring-files/#create-a-global-gitignore so I guess I am going to discard this :)

DanielRuf:
> Thanks for the PR! For this usually you can use a global gitignore https://help.github.com/articles/ignoring-files/#create-a-global-gitignore so I guess I am going to discard this :)

Sure but most users and projects do not mention this and this is not the default in general so most systems still report the .DS_Store files as new files.

DanielRuf:
In general: npm-debug.log would be also in a global .gitignore then ;-)

PR #440, opened 2018-05-02
chore: use npm ci

DanielRuf:


DanielRuf:
I close this PR as this would be slower and there would be no big benefits.

PR #450, opened 2018-05-16
add section under FAQ about usage with Rollup

Tomekmularczyk:
Related to #159. There are some changes around line 666, I hope it won't break anything. LOL

PR #452, opened 2018-05-25
Feature/update faq about using with rollup

Tomekmularczyk:
As proposed in PR #450 I create the same PR for `next` branch.

What I did is merged next branch to feature/update-FAQ-about-using-with-rollup and resolved conflicts which led to another merge commit. Not sure if that's okay from clean git history point of view.


giuseppeg:
we don't need the changes in `package.json`, probably you pulled

Tomekmularczyk:
@giuseppeg Somehow I didn't notice that package.json. Updated.

giuseppeg:
thanks applied the changes to the `next` branch right away

Issue #449, opened 2018-05-15
Unsafe transformation of className? Babel issue?


bathos:
#### Do you want to request a *feature* or report a *bug*?

Bug, but I’m unsure whether it’s a Babel 7 issue or a StyledJSX issue.

#### What is the current behavior?

After upgrading to Next 6, which entailed upgrading Babel to 7, the transpiled output for multiline className attributes that have been augmented by StyledJSX is no longer safely escaped and so produces syntax errors when parsed or evaluated.

#### If the current behavior is a bug, please provide the steps to reproduce and possibly a minimal demo or testcase in the form of a Next.js app, CodeSandbox URL or similar

```xml
<div
  jsx-attributes="
    can be multiline,
    like in xml and html
  "
  className="
    but the styledjsx babel transform does not seem
    to account for that when used with babel 7
  ">
  Module parse failed: Unterminated string constant (89:39)
  You may need an appropriate loader to handle this file type.
  |       lineNumber: 30
  |     },
  |     className: "jsx-104529230" + " " + "
  |     but the styledjsx babel transform does not seem
  |     to account for that when used with babel 7
</div>
```

Notice that the multiline attributes which is not processed by StyledJSX is still fine.

#### What is the expected behavior?

JSX string nodes not processed by StyledJSX are output safely. I would expect the same to be true for string nodes that are processed by StyledJSX. 

#### Environment (include versions)

* OS: OS X
* Browser: n/a
* styled-jsx (version): 2.2.6

#### Did this work in previous versions?

It did work fine prior to updating Next/Babel, with what I believe is the same version of StyledJSX (went from Next 5.1 to Next 6.0.2 but that dependency appears the same).

bathos:
Testing with other input, this seems like a broader conversion issue. For example, `className="\n\x21" outputs these characters unescaped, so that it becomes a string that is actually a newline and an exclamation point as if this were a JS string source and not a JSX string source.

giuseppeg:
Can you create a PR with a failing test? Just add an example to this file https://github.com/zeit/styled-jsx/blob/master/test/fixtures/attribute-generation-classname-rewriting.js

giuseppeg:
@bathos ping

giuseppeg:
@bathos can you reproduce with styled-jsx v3? I can't (maybe my editor removes those special characters)

bathos:
I added the requested test cases in the PR above, but it looks like it now has conflicts — I’ll have to try to bring it back in line w/ current master.

PR #456, opened 2018-06-21
two fixture elements in classname-rewriting that currently fail

bathos:
Per https://github.com/zeit/styled-jsx/issues/449#issuecomment-389419863


timneutkens:
Is this PR still relevant? Otherwise I think we should close it.

bathos:
I'm not sure if the issue is fixed or not. Last I checked the issue still existed but I might not have been on v3 yet. This PR was an attempt to add (failing, at the time) tests that demonstrate the bug I reported in #449; it was requested when I described the issue, but the PR got no action and went stale.

Even if the bug is fixed in v3, tests for something that was known to fail in the past might be considered valuable. On the other hand, I'm not super invested in it (not presently using Styled JSX day-to-day) and I'm reluctant to put the time into PR rehab given it didn't go anywhere last time. I'll close it.

PR #462, opened 2018-07-11
Feature/dynamic elements

twltwl:
Related issue https://github.com/zeit/styled-jsx/issues/147

Based on comment by @giuseppeg 

Automaticly scopes elements if possible ( works if declared inside of component )

for other cases a styled-jsx prop can be added

```javascript
// works
const Cmp = () => {
  const Element = 'div';
  return (<Element className="foo">
    <style jsx>{`.foo{...}`}</style>
  </Element>)
}

// wont work
const Element = 'div';
const Cmp = () => (<Element className="foo">
  <style jsx>{`.foo{...}`}</style>
</Element>)

// works
const Element = 'div';
const Cmp = () => (<Element className="foo" styled-jsx>
  <style jsx>{`.foo{...}`}</style>
</Element>)
```

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -41,11 +41,16 @@ export default function({ types: t }) {
           state.ignoreClosing = 0
         }
 
+        const tag = path.get('name')
+
         if (
           name &&
           name !== 'style' &&
           name !== STYLE_COMPONENT &&
           (name.charAt(0) !== name.charAt(0).toUpperCase() ||
+            Object.values(path.scope.bindings).some(binding =>
+              binding.referencePaths.some(r => r === tag)
+            ) ||
             (attributes &&
               Object.values(attributes).some(
                 ({ name, type }) =>
--- a/src/babel.js
+++ b/src/babel.js
@@ -26,7 +26,6 @@ export default function({ types: t }) {
       JSXOpeningElement(path, state) {
         const el = path.node
         const { name } = el.name || {}
-        const { attributes } = el || {}
 
         if (!state.hasJSXStyle) {
           return
@@ -50,12 +49,7 @@ export default function({ types: t }) {
           (name.charAt(0) !== name.charAt(0).toUpperCase() ||
             Object.values(path.scope.bindings).some(binding =>
               binding.referencePaths.some(r => r === tag)
-            ) ||
-            (attributes &&
-              Object.values(attributes).some(
-                ({ name, type }) =>
-                  type === 'JSXAttribute' && name.name === 'styled-jsx'
-              )))
+            ))
         ) {
           if (state.className) {
             addClassName(path, state.className)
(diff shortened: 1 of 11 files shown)


src/babel.js
```diff
@@ -40,11 +41,21 @@ export default function({ types: t }) {
...
         if (
           name &&
           name !== 'style' &&
           name !== STYLE_COMPONENT &&
-          name.charAt(0) !== name.charAt(0).toUpperCase()
+          (name.charAt(0) !== name.charAt(0).toUpperCase() ||
+            Object.values(path.scope.bindings).some(binding =>
+              binding.referencePaths.some(r => r === tag)
+            ) ||
+            (attributes &&
```
giuseppeg: can we remove this feature for now?

test/fixtures/attribute-generation-classname-rewriting.js
```diff
@@ -34,6 +36,9 @@ export default () => (
     <div {...props} {...rest} />
     <div {...props} data-foo {...rest} />
     <div {...props} className={'test'} data-foo {...rest} />
+    <Element styled-jsx />
```
giuseppeg: I won't want to include this feature here for now, let's fix one thing at the time

test/fixtures/dynamic-element.js
```diff
@@ -0,0 +1,14 @@
...
+    <Element className="root">
+      <p>dynamic element</p>
+      <style jsx>{`
+        .root {
+          background: red;
+        }
+      `}</style>
+    </Element>
+  )
+}
```
giuseppeg: awesome can you add a few more tests cases? i.e. a few more combinations to where `Element` is defined outside of the component, and maybe when it is lowercase too (any edge case you can think of). Also can you add a class component too?

src/babel.js
twltwl: Sure, if we want to add it at a later point let me know

test/fixtures/dynamic-element.js
twltwl: Maybe Im missing something, but I dont think it will work if Element is defined outside of the component. ( Thats why I added the styled-jsx option )
I´ll add the testcases

giuseppeg: correct, it won't work (it won't add the className) and we want a snapshot of that! ;)

twltwl: 👍

giuseppeg:
@twltwl this branch is good to merge, can you just remove node `6` from travis https://github.com/zeit/styled-jsx/blob/a55c821a998b9a7210c03a30c4ae33de7ec94eb6/.travis.yml#L3 ? and maybe add a new line for node `10`?

giuseppeg:
@twltwl thanks so much for this great contribution! 🙌

iamawebgeek:
@giuseppeg I cannot use this feature for some reason. Babel is not stripping out the `styled-jsx` attribute and adding a scoped `className`. How can I figure out what is the problem?

PR #464, opened 2018-07-18
Switch to async safe lifecycle methods

### LINKED ISSUE #461: Warnings in AsyncMode
Hello,

If you wrap an app using styled-jsx with an `AsyncMode` component (`React.unstable_AsyncMode`) a warning is reported to the console : 

```
Warning: Unsafe lifecycle methods were found within a strict-mode tree:
componentWillMount: Please update the following components to use componentDidMount instead: JSXStyle
componentWillUpdate: Please update the following components to use componentDidUpdate instead: JSXStyle
```

We can trace those two deprecated lifecycles calls to `src/style.js`.

- For `componentWillMount`, I think we could just move the `styleSheetRegistry.add(this.props)` call to the constructor of the `JSXStyle` component.

- For `componentWillUpdate`, the decision is trickier because of the react peerDependency that could be react 15... If it's dropped, the call to `styleSheetRegistry.update(this.props, nextProps)` could be moved to (weird) `getSnapshotBeforeUpdate` to prevent FOUC instead of `componentDidUpdate`.

What do you think ?

giuseppeg:
Fixes #461 #457


giuseppeg:
should we remove `15.x.x` from the `peerDependencies`?

test/browser.js
```diff
@@ -1,68 +0,0 @@
-import test from 'ava'
```
giuseppeg: I removed this test because it runs in jsdom and doesn't work with React 16. 
Someday we'll need to setup proper e2e tests. I don't have the time for it now though

PR #466, opened 2018-07-19
Add webpack loader for external CSS files

giuseppeg:
In v3 (which we will likely release when I am back from holidays) I improved how external styles work and as a result of that now with a loader like this we support external `.css` files. This means that you can write your styles in a regular `.css` file and benefit from all the tooling (including preprocessors, autocompletion, syntax highlighting) and then consume them with styled-jsx.


giuseppeg:
> Actually, not a fan of CSS comment but no better idea.

Same but this is just a last resort for one-offs. Usually you could use the loader twice to match `.global.css` and `.css`.

giuseppeg:
@nkzawa what do you think about using the `MiniCssExtractPlugin` pattern. They expose a webpack plugin and the `loader` is a static property in the plugin class. For now we could expose a dummy (empty class) with this static property but in future the plugin could do static extraction. Alternatively I could rename `styled-jsx/webpack` to `styled-jsx/webpack-loader`

PR #473, opened 2018-08-07
Readme: Update Atom instructions with support for css`...`

rstacruz:
The Atom snippet only works for `<styled jsx>{...}` snippets. This adds support for `` css`...` `` as well.

Here's how I have it configured (pardon the font ligatures):

image

I also have it configured for `source.css.postcss` since I often use styled-jsx-postcss, but I didn't note that in this PR.


rstacruz:
Here's how it looks like:

image

giuseppeg:
awesome thank you!

Issue #474, opened 2018-08-08
StyleSheetRegistry in optimizeForSpeed mode should keep track of invalid rules


rgbkrk:
#### Do you want to request a *feature* or report a *bug*?

Feature (to report on user level issues)

#### What is the current behavior?

Same as #319 (which is closed and locked), there's a persistent issue we're facing when developing component libraries using styled-jsx that later get used in next.js apps. If the browser rejects the sheet (with a warning), styled jsx doesn't notice and later causes an error when trying to delete that style sheet.

![image](https://user-images.githubusercontent.com/836375/43859413-62a983a4-9b05-11e8-9188-4fce68784161.png)

![image](https://user-images.githubusercontent.com/836375/43859417-6645ead4-9b05-11e8-820d-c45dba1be92e.png)

#### If the current behavior is a bug, please provide the steps to reproduce and possibly a minimal demo or testcase in the form of a Next.js app, CodeSandbox URL or similar

Example to reproduce with, along with steps: https://github.com/nteract/nteract/pull/3212

There is always the chance that the optimizer is just _wrong_ and isn't validating it's output! Not sure where that happens though.

#### What is the expected behavior?

What I'd _love_ is for styled-jsx to be able to detect when a stylesheet is bad _in advance_. I can't tell from how styled-jsx operates when something is going to be wrong as a contributor is adding it. I need tooling to let us know what to fix / change.

#### Environment (include versions)

* OS: macOS
* Browser: Chrome
* styled-jsx (version): 2.2.6

#### Did this work in previous versions?

¯\\\_(ツ)\_/¯

rgbkrk:
One thing that I know we're doing for our built components is that we set `NODE_ENV=production` while running the babel build. That affects styled-jsx as far as I'm aware.

giuseppeg:
it seems that you have source maps in production and those don't work when `optimizeForSpeed` is `true`: https://github.com/zeit/styled-jsx#optimizeforspeed let me know if this helps (set it to `false` explicitly)

rgbkrk:
We are currently setting `optimizeForSpeed` to false explicitly. I'm a bit surprised about the source maps being enabled in production, we're not enabling them in our next config. How do we detect this problem in advance at build time? When modifying webpack config for next, does this have to be explicitly set somehow?

giuseppeg:
you can pass styled-jsx options to Next as explained here https://github.com/zeit/next.js/#customizing-babel-config

[...]

PR #476, opened 2018-08-16
register invalid rules for removal

rgbkrk:
Attempts to handle fix mentioned in https://github.com/zeit/styled-jsx/issues/474#issuecomment-413243523

I'm not sure this is exactly the behavior wanted since it's storing an empty string when the rule is invalid (while not keeping the invalid stylesheet).

### DIFF
--- a/test/stylesheet-registry.js
+++ b/test/stylesheet-registry.js
@@ -92,6 +92,23 @@ test(
   })
 )
 
+test(
+  'it does not throw when inserting an invalid rule',
+  withMock(withMockDocument, t => {
+    const registry = makeRegistry()
+
+    // Insert a valid rule
+    registry.add({ styleId: '123', css: [cssRule] })
+
+    t.notThrows(() => {
+      // Insert an invalid rule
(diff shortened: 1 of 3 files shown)


test/stylesheet-registry.js
```diff
@@ -81,6 +81,7 @@ test('add - filters out invalid rules (index `-1`)', t => {
 
   t.deepEqual(registry.cssRules(), [
     ['jsx-123', 'div { color: red }'],
+    ['jsx-456', ''],
```
giuseppeg: we shouldn't have empty rules, probably we need to filter them out.

giuseppeg: https://github.com/avajs/ava#notthrowsfn-message

rgbkrk:
I'd love for you to take it over.  🤗  Thank you.

giuseppeg:
@rgbkrk thanks for the contribution! I will release this weekend or next week

rgbkrk:
Thank you!

Issue #423, opened 2018-03-21
[feature] Content Security Policy (CSP)


thealjey:
It is possible to specify a `nonce` prop when rendering `<NextScript />`, but there's currently no way to do the same for `styled-jsx` inline styles, forcing me to resort to using `'unsafe-inline'` for styles.

I would like to be able to pass any number of props supported by `<style />` to `'styled-jsx/server'` and have them rendered in the DOM.

I can make a pull request (although, that would require 2 pull requests to both `next` and `styled-jsx`), if you are open to the idea.

giuseppeg:
Definitely open to add this if you prepare a PR.

jlecomte:
Here is a workaround:

```js
let styles = flush()
styles = styles.map(({props, key}) =>
  React.createElement('style', {...props, key, nonce})
)
```

I purposefully did not use JSX to avoid going through the styled-jsx babel plugin, which causes all kinds of strange issues.

thealjey:
Unfortunately, I will most probably not have the time in the next several months to do these pull requests.
Maybe someone else will do them in the meantime 🤞

andy-viv:
@thealjey: the workaround by @jlecomte would be in your code, not in styled-jsx's code.

thealjey:
@andy-viv 
1. I was not referring to his workaround
2. I would rather not have to do any workarounds
3. Next.js does the `flush`, so more workarounds are necessary to make this work

giuseppeg:
fixed in #482

PR #482, opened 2018-08-28
Support Strict CSP with a nonce

dav-is:
### Content Security Policy

Strict CSP is supported. You must pass a nonce as a parameter to either `flush(nonce)` or `flushToHTML(nonce)` **and** set a `<meta property="csp-nonce" content="nonce">` tag.

You should generate a nonce per request.
```jsx
import nanoid from 'nanoid'

const nonce = Buffer.from(nanoid()).toString('base64') //ex: N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3
```

Your CSP policy must have this same nonce as well.
`Content-Security-Policy: default-src 'self'; style-src 'self' 'nonce-N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3';`

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -387,16 +387,18 @@ duplicate styles are avoided.
 
 ### Content Security Policy
 
-Strict CSP is supported. You must pass a nonce as a parameter to either `flush(nonce)` or `flushToHTML(nonce)` **and** set a `<meta property="csp-nonce" content="nonce">` tag.
+Strict [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) is supported. 
 
-You should generate a nonce per request.
-```jsx
+You should generate a nonce **per request**.
+```js
 import nanoid from 'nanoid'
 
 const nonce = Buffer.from(nanoid()).toString('base64') //ex: N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3
 ```
 
-Your CSP policy must have this same nonce as well.
+You must then pass a nonce to either `flush({ nonce })` or `flushToHTML({ nonce })` **and** set a `<meta property="csp-nonce" content={nonce} />` tag.
+
+Your CSP policy must share the same nonce as well (the header nonce needs to match the html nonce and remain unpredictable).
 `Content-Security-Policy: default-src 'self'; style-src 'self' 'nonce-N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3';`
 
 ### External CSS and styles outside of the component
--- a/readme.md
+++ b/readme.md
@@ -391,9 +391,9 @@ Strict CSP is supported. You must pass a nonce as a parameter to either `flush(n
 
 You should generate a nonce per request.
 ```jsx
-import uuidv4 from 'uuid/v4'
+import nanoid from 'nanoid'
 
-const nonce = new Buffer(uuidv4()).toString('base64') //ex: N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3
+const nonce = Buffer.from(nanoid()).toString('base64') //ex: N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3
 ```
 
 Your CSP policy must have this same nonce as well.
--- a/readme.md
+++ b/readme.md
@@ -396,7 +396,7 @@ import nanoid from 'nanoid'
 const nonce = Buffer.from(nanoid()).toString('base64') //ex: N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3
 ```
 
-You must then pass a nonce to either `flush({ nonce })` or `flushToHTML({ nonce })` **and** set a `<meta property="csp-nonce" content={nonce} />` tag.
+You must then pass a nonce to either `flushToReact({ nonce })` or `flushToHTML({ nonce })` **and** set a `<meta property="csp-nonce" content={nonce} />` tag.
 
 Your CSP policy must share the same nonce as well (the header nonce needs to match the html nonce and remain unpredictable).
 `Content-Security-Policy: default-src 'self'; style-src 'self' 'nonce-N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3';`
--- a/readme.md
+++ b/readme.md
@@ -385,6 +385,20 @@ It's **paramount** that you use one of these two functions so that
 the generated styles can be diffed when the client loads and
 duplicate styles are avoided.
 
+### Content Security Policy
+
+Strict CSP is supported. You must pass a nonce as a parameter to either `flush(nonce)` or `flushToHTML(nonce)` **and** set a `<meta property="csp-nonce" content="nonce">` tag.
+
+You should generate a nonce per request.
+```jsx
+import uuidv4 from 'uuid/v4'
+
+const nonce = new Buffer(uuidv4()).toString('base64') //ex: N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3
+```
+
+Your CSP policy must have this same nonce as well.
+`Content-Security-Policy: default-src 'self'; style-src 'self' 'nonce-N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3';`
+
 ### External CSS and styles outside of the component
 
 In styled-jsx styles can be defined outside of the component's render method or in separate JavaScript modules using the `styled-jsx/css` library. `styled-jsx/css` exports three tags that can be used to tag your styles:
--- a/src/lib/stylesheet.js
+++ b/src/lib/stylesheet.js
@@ -29,12 +29,11 @@ export default class StyleSheet {
     this._rulesCount = 0
 
     const node =
-      this._isBrowser &&
-      document.querySelector('meta[property="csp-nonce"]')
+      this._isBrowser && document.querySelector('meta[property="csp-nonce"]')
     this._nonce = node ? node.getAttribute('content') : null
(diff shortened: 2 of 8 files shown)


src/server.js
```diff
@@ -1,26 +1,29 @@
...
+      nonce,
       dangerouslySetInnerHTML: {
         __html: css
       }
     })
   })
 }
 
-export function flushToHTML() {
+export function flushToHTML(nonce = undefined) {
```
giuseppeg: `= undefined` is redundant, you can remove it.

giuseppeg: can we make the arguments an object? This will allow us to add more options in the future.

```js
function flushToHTML(options = {}) {
  //   options.nonce 
}
```

src/lib/stylesheet.js
```diff
@@ -29,6 +27,11 @@ export default class StyleSheet {
     this._tags = []
     this._injected = false
     this._rulesCount = 0
+
+    const node =
+      typeof document === 'object' &&
```
giuseppeg: Here you can use `this._isBrowser`

readme.md
```diff
@@ -385,6 +385,20 @@ It's **paramount** that you use one of these two functions so that
 the generated styles can be diffed when the client loads and
 duplicate styles are avoided.
 
+### Content Security Policy
+
+Strict CSP is supported. You must pass a nonce as a parameter to either `flush(nonce)` or `flushToHTML(nonce)` **and** set a `<meta property="csp-nonce" content="nonce">` tag.
```
giuseppeg: I am not super familiar with CSP but can you have multiple `csp-nonce` meta tags in the same page?

readme.md
```diff
@@ -385,6 +385,20 @@ It's **paramount** that you use one of these two functions so that
...
+
+You should generate a nonce per request.
+```jsx
```
giuseppeg: this can be `js`

readme.md
dav-is: The `csp-nonce` meta tag is a convention, not a standard, so there really shouldn't be any other common uses of the tag. This is kind of a workaround that allows dynamic styling to work with CSP. The CSP Policy itself is defined in the header.

src/lib/stylesheet.js
dav-is: I feel like changing this might break [my test](https://github.com/zeit/styled-jsx/pull/482/files#diff-be6d9079ddafc0af10f17b86a4033dcaR109) :worried: because there isn't any browser testing implemented in this repo :frowning_face:

giuseppeg: probably in your test you can do `global.window = true`.

giuseppeg: ah no you can pass `isBrowser: true` to `makeRegistry` https://github.com/zeit/styled-jsx/blob/master/test/stylesheet-registry.js#L8-L12

test/stylesheet-registry.js
```diff
@@ -104,6 +104,26 @@ test('add - sanitizes dynamic CSS on the server', t => {
   ])
 })
 
+test('add - nonce is properly fetched from meta tag', t => {
+  // We need to stub a document in order to simulate the meta tag
+  global.document = {
```
giuseppeg: now that I look at this again I realize that we should restore `global.document` after the test:

```js
const originalDocument = global.document

global.document = { /* ... */ }

// test ...

global.document = originalDocument
```

src/lib/stylesheet.js
dav-is: @giuseppeg Will this not cause unintended side effects because we are telling the application it is running in a browser and it's not really? If somewhere else in the application it checks `isBrowser` and if I set it to true, it might try to execute some other type of browser only script.

test/stylesheet-registry.js
dav-is: It's undefined in the test environment as far as I could tell.

giuseppeg: yea but better reset in case something else relies on it to not be defined (good practice)

src/lib/stylesheet.js
giuseppeg: this is an isolated unit test so it is perfectly fine. If you are talking about  `src/lib/stylesheet.js` instead that's fine too since you are checking that `typeof document === 'object'` is defined i.e. you are in a browser environment.

dav-is: I was right about the unintended side effects. It's possible that `isBrowser == true` and `typeof document === 'undefined'`, which will create an error.

giuseppeg: it should behave the same when isBrowser is true. If there are failing tests this is a good thing and we should look into this issue. In a real browser `typeof document === 'object'` will be the same as using `this._isBrowser` because the latter is just `typeof window !== 'undefined;`

src/lib/stylesheet.js
```diff
@@ -41,7 +44,7 @@ export default class StyleSheet {
       this._rulesCount === 0,
       'optimizeForSpeed cannot be when rules have already been inserted'
     )
-    this.flush()
+    this.flush(options.flush)
```
giuseppeg: why do you need to pass `options.flush`?

src/stylesheet-registry.js
```diff
@@ -94,8 +94,8 @@ export default class StyleSheetRegistry {
     this.remove(props)
   }
 
-  flush() {
-    this._sheet.flush()
+  flush(options = {}) {
```
giuseppeg: what is this for?

src/lib/stylesheet.js
dav-is: If you want me to break the tests, I will, I just don't have time to fix them myself.

src/lib/stylesheet.js
dav-is: In case that method gets other options eventually.

src/stylesheet-registry.js
dav-is: In case someone is using this method and needs to set the nonce.

giuseppeg: I don't think it is necessary also `StyleSheet#flush` is not using those options https://github.com/zeit/styled-jsx/blob/368a6e4e3e1211e84b6bc4bb4db44dcd426a0bd9/src/lib/stylesheet.js#L192

Can you revert this change?

giuseppeg:
Thanks for your patience @dav-is! To fix the tests you can create a little helper function to mock `document` and `document.querySelector` like you did in your test. Something like

```js
function withMock(mockFn, testFn) {
  const cleanUp = mockFn()
  if (typeof cleanUp !== 'function') {
    throw new Error('mockFn should return a cleanup function')
  }
  testFn()
  cleanUp()
}

function withMockDocument() {
  const originalDocument = global.document
  // We need to stub a document in order to simulate the meta tag
  global.document = {
    querySelector(query) {
      t.is(query, 'meta[property="csp-nonce"]')
      return {
        getAttribute(attr) {
          t.is(attr, 'content')
          return 'test-nonce'
        }
      }
    }
  }

  return () => {
    global.document = originalDocument
  }
}


withMock(
  withMockDocument,
  () => {
    const registry = makeRegistry()
    registry.add({ styleId: '123', css: [cssRule] })
    t.is(registry._sheet._nonce, 'test-nonce')
  }
)
```

dav-is:
@rauchg All done :)

dav-is:
@giuseppeg All done :)

giuseppeg:
@thealjey do you mind taking a quick look at this PR and let us know if you had something like this in mind when you opened #423 ?

timneutkens:
@thealjey sorry for the second ping, could you have a look? 🙏 🙌

thealjey:
oops, sure! will do in a couple hours :)

thealjey:
The API is exactly as I thought it should be (although, this will require the changes to be also made to Next.js to actually pass nonce to `flush` and, preferably, create the meta tag also).
I do have a question regarding the actual changes.
The readme states that you can pass `nonce` to the `flush` function, but I do not see `flush` actually accept it (`flushToHTML` and `flushToReact` both do, but not `flush`). It only reads the value from the meta tag on the client.
Maybe I just don't see something.

readme.md
```diff
@@ -385,6 +385,22 @@ It's **paramount** that you use one of these two functions so that
...
+Strict [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) is supported. 
+
+You should generate a nonce **per request**.
+```js
+import nanoid from 'nanoid'
+
+const nonce = Buffer.from(nanoid()).toString('base64') //ex: N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3
+```
+
+You must then pass a nonce to either `flush({ nonce })` or `flushToHTML({ nonce })` **and** set a `<meta property="csp-nonce" content={nonce} />` tag.
```
giuseppeg: we should replace `flush` with `flushToReact`

giuseppeg:
@thealjey thank you <3

> The readme states that you can pass nonce to the flush function, but I do not see flush actually accept it 

yep we should fix tht

dav-is:
@timneutkens @giuseppeg Do you think we could get a release?

dav-is:
@giuseppeg Could you please make a new release?

giuseppeg:
@dav-is published 3.1.0 sorry for the delay

xereda:
The updated documentation states that we should apply registry.styles({ nonce }). Where and how can I do this in a Next.js project (14.+)?

dav-is:
@xereda you can follow [this guide](https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy#adding-a-nonce-with-middleware). I believe adding the CSP header to the *request* headers allows Next.js to automatically use the nonce. Then the client uses the CSP additionally sent in the *response* headers.

I would also verify the `<meta property="csp-nonce" content={nonce} />` is returned in the html, if not, you may have to do it manually using the `x-nonce` request header mentioned in the guide.

xereda:
@dav-is I did all of this. I was even able to apply it to scripts, but it's not working for styles. I have no idea where to implement "registry.styles({ nonce })".

PR #483, opened 2018-08-31
Render styles before content

timneutkens:
On zeit.co we noticed strange jumps in content that uses `transition: all` because styles got applied after the content was rendered to the screen.


timneutkens:
Solved by #483

PR #485, opened 2018-08-31
Move server-only method to server file

timneutkens:
It seems like this method was only used in the server flusher, so it's not needed to ship this extra code to the client side.


timneutkens:
Let's merge it for styled-jsx v4 👍

jamestalmage:
I also removed in #563. Since we're abandoning the Singleton instance, this method isn't even possible anymore.

PR #493, opened 2018-09-20
Add styled-jsx/babel-test plugin

### LINKED ISSUE #459: Making styled-jsx work better with Jest snapshots
### Do you want to request a *feature* or report a *bug*?

A feature.

### What is the current behavior?

Most people don't want to include tons of `jsx-******` classnames in their Jest snapshots, nor lots of vendor prefixed CSS properties. It just makes the snapshot harder to read, and provides no useful information. There's no need to test styled-jsx's functionality every time you test your own components.

To make things worse, whenever you change your CSS, you end up with a lot of diff noise as all your DOM nodes get new classnames. All you really need to see in the diff is the lines of CSS that changed.

**Current best workaround**

As outlined in this [blog post](https://kevinjalbert.com/jest-snapshots-reducing-styled-jsx-noise/), you can simply switch off the `styled-jsx/babel` plugin for the `test` [environment](https://babeljs.io/docs/en/babelrc/#env-option) in your .babelrc. So your snapshots end up containing uncompiled `<style jsx>...</style>` elements, which is what you want.

**Problems with the workaround**

When rendering with enzyme or react-test-renderer, you end up with a tons of `console.error` noise coming from React when running your tests:

```
Warning: Received `true` for a non-boolean attribute `jsx`.
(...followed by a long stack)
```

Another problem is, when using something like enzyme's `wrapper.text()` to make an assertion about the rendered text content of your component, you get the component's raw CSS concatenated onto the end of the real text content, because the content of a `<style>` tag technically counts as a text node. (You can avoid this by selecting a descendant instead of the wrapper, but that sometimes means you have to add extra structural elements purely to work around this issue, which isn't ideal.)

### What is the expected behavior?

I suggest exposing an alternative Babel plugin entrypoint at `"styled-jsx/babel-test"` — or alternatively adding some kind of 'test mode' option on the regular `"styled-jsx/babel"` plugin to use this behaviour.

This would just do a light-touch conversion of `<style jsx>...</style>` elements, just enough to make them suitable for snapshotting. Like this:

```jsx
<div className="foo">
  <style jsx>{`
     .foo { transition: all 1s ease; }
  `}</style>
</div>

// ...would compile to something like:
<div className="foo">
  <style data-styled-jsx="{`
     .foo { transition: all 1s ease; }
  `} />
</div>
```

- The plain `jsx` boolean attribute is removed to prevent React warnings.
- The raw CSS should not be prefixed/modified in any way.
- There should be no `jsx-*****` classnames generated anywhere, so the rest of the component remains untouched.
- The CSS is moved into a data attribute so it doesn't count as a text node.

callumlocke:
Adds `styled-jsx/babel-test` plugin as per discussion with @giuseppeg.

The readme entry feels a bit long, but I think it's necessary to give some background to explain why you might (or might not) want to use it.

Closes #459.

### DIFF
--- a/babel-test.js
+++ b/babel-test.js
@@ -1,0 +1,1 @@
+module.exports = require('./dist/babel-test')
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "lib",
     "server.js",
     "babel.js",
+    "babel-test.js",
     "style.js",
     "macro.js",
     "css.js",
--- a/readme.md
+++ b/readme.md
@@ -35,6 +35,7 @@ Code and docs are for v3 which we highly recommend you to try. Looking for style
 - [CSS Preprocessing via Plugins](#css-preprocessing-via-plugins)
   * [Plugin options](#plugin-options)
   * [Example plugins](#example-plugins)
+- [Rendering in tests](#rendering-in-tests)
 - [FAQ](#faq)
   * [Warning: unknown `jsx` prop on &lt;style&gt; tag](#warning-unknown-jsx-prop-on-style-tag)
   * [Can I return an array of components when using React 16?](#can-i-return-an-array-of-components-when-using-react-16)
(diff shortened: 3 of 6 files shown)


giuseppeg:
@callumlocke is it ok if we release next week? or do you need this asap?

callumlocke:
next week is good 👍

Issue #509, opened 2018-10-24
Plugins Options: `filename` is `undefined`


ryanfitzer:
#### Do you want to request a *feature* or report a *bug*?

Bug.

This is due to the wrong property name being used in [_utils.js on line 550](https://github.com/zeit/styled-jsx/blob/master/src/_utils.js#L550). Instead of `fileInfo.filename`, it should be `fileInfo.sourceFileName`.

#### What is the current behavior?

The `filename` property is `undefined` when passed to plugins. This was noticed while using `styled-jsx-plugin-postcss`.

#### If the current behavior is a bug, please provide the steps to reproduce and possibly a minimal demo or testcase in the form of a Next.js app, CodeSandbox URL or similar

Inspect the `options` object passed to [plugins](https://github.com/zeit/styled-jsx/blob/master/readme.md#plugin-options). The result is always:

```
{
  babel: {
    ...
    filename: undefined
    ...
  }
}
```

#### What is the expected behavior?

The `filename` property should be defined.

#### Environment (include versions)

* OS: macOS 10.12.6
* Browser: N/A
* styled-jsx (version): 3.1.0

#### Did this work in previous versions?

N/A

ryanfitzer:
PR #510 opened to fix this issue.

PR #510, opened 2018-10-24
Fix typo explained in #509

ryanfitzer:
Fix typo explained in #509.

### DIFF
--- a/src/_utils.js
+++ b/src/_utils.js
@@ -547,7 +547,7 @@ export const processCss = (stylesInfo, options) => {
     vendorPrefixes,
     sourceMaps: useSourceMaps,
     isGlobal,
-    filename: fileInfo.filename
+    filename: fileInfo.sourceFileName
   }
 
   let transformedCss


giuseppeg:
to be honest I don't know exactly what is the difference between filename and sourceFileName:

https://github.com/zeit/styled-jsx/blob/1b3497517ec8743db3f236fe3191482578da121b/src/babel.js#L230-L238

I asked on the Babel slack community, let's wait for a reply.

The part you changed could be correct and the bug could be here https://github.com/zeit/styled-jsx/blob/1b3497517ec8743db3f236fe3191482578da121b/src/babel-external.js#L205-L209 (which is not computing and passing `filename`

giuseppeg:
@ryanfitzer do you mind taking a look at these and see if `sourceFileName` is really what we need? I won't have time tonight/in the following days (maybe on Wed)

https://babeljs.io/docs/en/options#filename
https://babeljs.io/docs/en/options#filenamerelative
https://babeljs.io/docs/en/options#sourcefilename

ryanfitzer:
@lluisrojass If you get time, can you take a look at this before Wednesday? I won't have time until Monday.

giuseppeg:
closed in favor of #526

Issue #480, opened 2018-08-22
With last update I am now getting lots of logging during the build.


richtera:
👋 friend. Welcome to styled-jsx and thanks for contributing!

⚠️ IMPORTANT ⚠️

If you need help or have a question about styled-jsx please ask it on Spectrum https://spectrum.chat/styled-jsx or join our Slack community at https://zeit.chat and ask it in the `#next` channel.

Before you open a new issue please take a look at our [**Frequent Asked Questions**](https://github.com/zeit/styled-jsx#faq) and [**open issues**](https://github.com/zeit/styled-jsx/issues).

Remember, often time asking in chat or looking at the FAQ/issues can be *faster than waiting for us to reply to a new issue**.

If you are here to report a bug or request a feature please remove this introductory section and fill out the information below!

---

#### Do you want to request a *feature* or report a *bug*?
Small bug

#### What is the current behavior?
During build I get notifications of plugins loaded:
```
[styled-jsx] Loading plugin from path: styled-jsx-plugin-sass
```

#### If the current behavior is a bug, please provide the steps to reproduce and possibly a minimal demo or testcase in the form of a Next.js app, CodeSandbox URL or similar

1. Use styled-jsx with the sass plugin.
2.

#### What is the expected behavior?
No additional logging

#### Environment (include versions)

* OS: Mac OS node 9.11
* Browser: (this is a build problem)
* styled-jsx (version):
```
├─ next@6.1.1
│  └─ styled-jsx@2.2.6
└─ styled-jsx@3.0.2
```

#### Did this work in previous versions?
Yes worked fine prior to the last update and it's not a fatal issue, but kind of annoying.

giuseppeg:
That's expected. Is it logged many times or just one?

richtera:
I now get about 20 of this messages when I start dev or build

giuseppeg:
ouch! Can you make a simple test repo that I can clone and try quickly? If there is a bug it should be fairly easy to fix

richtera:
It's just a nextjs app. I can put together a demo repo next week I think. But as far as I know I am not doing anything custom with styled-jsx except extracting the css in one place to be able to send a css blob to my cms.

giuseppeg:
yeah but I don't have a lot of free time recently. So it'd be great if you could set up a repo and integrate a styled-jsx plugin and add a few styles so that I can clone and start to debug right away.

richtera:
[...]

PR #511, opened 2018-10-25
Enable logging only development

meabed:
When you configure plugins for `styled-jsx` you get huge amount of logs with every jsx tag.
so a quick solution is to allow this log only on development ENV, or other ideas are welcome :) 
Currently it so annoying to get this logs in the build!

### DIFF
--- a/src/_utils.js
+++ b/src/_utils.js
@@ -642,5 +642,5 @@ export const setStateOptions = state => {
 }
 
 export function log(message) {
-  console.log('[styled-jsx] ' + message)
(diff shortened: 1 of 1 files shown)


giuseppeg:
Sorry I'd rather fix this properly. Here is the reason why this prints the logging info so many times https://github.com/zeit/styled-jsx/issues/480#issuecomment-432541224 would you try to implement that fix instead?

meabed:
@giuseppeg this is babel visitor for AST traversal. and it doesn't share global state between all files as every file could have different syntax tree and plugins etc...

Issue #521, opened 2018-11-10
Type function in webpack config executes only once


danielr18:
This is a snippet of the webpack loader:

```js
module.exports = function(content) {
  // ...
  const options = loaderUtils.getOptions(this) || {}
  // ...
  // Calls type with the current file name.
  if (typeof options.type === 'function') {
    options.type = options.type(this.resourcePath)
  }
  // ...
}
```

The options object returned by `loaderUtils.getOptions` should be treated as read-only, as it's used across multiple invocations. When `options.type` is a function, it will set its result back to `options.type`,  this causes that in the next invocation it will still have the result of the function that was called with the previous filename, not a function anymore.

danielr18:
I created a temporary loader that appends a comment with the type, so that I can still use type function until it gets fixed.

```js
// File: styled-jsx-type-loader.js
const loaderUtils = require('loader-utils')

module.exports = function(source) {
  this.cacheable()
  const options = Object.assign({}, loaderUtils.getOptions(this))
  if (typeof options.type === 'function') {
    const type = options.type(this.resourcePath)
    const prefix = `/* @styled-jsx=${type} */`
    source = prefix + '\n' + source
  }
  this.callback(null, source)
}
```
To use it, just add it after the `styled-jsx` one:

```js
{
  loader: require('styled-jsx/webpack').loader,  
},
{
  loader: require.resolve('path to styled-jsx-type-loader'),
  options: {
    type: filename => {}
  }
}
```

giuseppeg:
just release `3.1.2` with a fix

PR #522, opened 2018-11-10
Fix options type function to always execute

danielr18:
Solves #521.


giuseppeg:
good catch thanks for the PR!

Issue #539, opened 2019-01-15
Issue when using Spread Operator in JSX


samholmes:
There seems to be an issue with spread props and style-jsx:

```jsx
const Home = () => (
  <div>
    <div {...{ id: 'foo' }}>test</div>

    <style jsx>{`
      #foo {
        background: blue;
      }
    `}</style>
  </div>
)

export default Home
```

Gives an error:

```
./pages/index.js
Module build failed (from ./node_modules/next/dist/build/webpack/loaders/next-babel-loader.js):
TypeError: Cannot read property '0' of null
```

The only fix for this bug is to remove the `<style jsx>` element altogether.

Related: https://github.com/zeit/next.js/issues/6057

giuseppeg:
Yeah this is an edge case. Would you like to make a PR?

samholmes:
I’d love to however I’m not familiar enough with the code base to begin to understand what the issue could be. Any thoughts?

doniyor2109:
@giuseppeg Hi. Can I take this issue?

AnatoleLucet:
I found something that is probably related.

When you spread an object returned by a function, the prop `className` is overwritten.

I made an example:
https://codesandbox.io/s/nifty-jennings-dl164?file=/pages/index.js

If you inspect the div, you'll see that it doesn't have the class `my-class`.

PR #541, opened 2019-01-19
Handle jsx spread attribute with object expression

doniyor2109:
It should solve #539


src/_utils.js
```diff
@@ -28,22 +28,41 @@ export const addClassName = (path, jsxId) => {
...
+
+        if (~index) {
+          className = attr.get('argument').get(`properties.${index}`)
+          break
+        }
+      }
+
+      if (
+        t.isMemberExpression(node.argument) ||
+        t.isIdentifier(node.argument)
```
giuseppeg: is this if necessary? what else can we have here? maybe you can add a `continue` inside of `if (t.isObjectExpression(node.argument)) {`

test/fixtures/spread-operator.js
```diff
@@ -0,0 +1,23 @@
+const Home = () => (
```
giuseppeg: can you remove this fixture file and add a few cases to this https://github.com/zeit/styled-jsx/blob/803896fc36f81d6689ee0f074306fc130a731698/test/fixtures/attribute-generation-classname-rewriting.js ?

```jsx
<div {...{ id: 'foo' }} />
<div {...{ className: 'foo' }} />
<div {...{ className: 'foo' }} className="test" />
<div className="test" {...{ className: 'foo' }} />
<div  {...{ className: 'foo' }} {...bar} />
<div  {...{ className: 'foo' }} {...bar} className="test" />
<div className="test"  {...{ className: 'foo' }} {...bar} />
// more corner cases???
```

test/fixtures/attribute-generation-classname-rewriting.js
```diff
@@ -36,6 +36,10 @@ export default () => {
       <div {...props} {...rest} />
       <div {...props} data-foo {...rest} />
       <div {...props} className={'test'} data-foo {...rest} />
+      <div {...{ className: 'test' }} />
+      <div {...{ className: 'test' }} data-test="test" />
+      <div {...{ className: 'test' }} {...props} data-test="test" />
+      <div {...props} {...{ id: 'test' }} />
```
giuseppeg: thanks can you add the other tests form the comment? It is important to make sure that the element gets a `className` prop the spread is merged correctly

doniyor2109: My bad, I did not see your tests cases 🙈

doniyor2109: I have updated test cases. 
Can you verify that it handles multiple classNames correctly?

src/_utils.js
doniyor2109: > what else can we have here?

You mean other conditions?

giuseppeg: yea I mean we can spread:

* ObjectExpression
* MemberExpression
* Identifier

so I was thinking that you can add a `continue` inside of the `isObjectExpression` if statement body, this way you don't need this other if statement.

Probably there are other weird edge cases like an inline function call like `{...foo()}` but shrug :P otherwise your code is fine if we want to keep the safety check :) Your call I am fine either ways ;)

test/babel6/snapshots/attribute.js.md
```diff
@@ -146,6 +146,13 @@ Generated by [AVA](https://ava.li).
           <div {...props} {...rest} className={"jsx-2886504620" + " " + (rest.className != null && rest.className || props.className != null && props.className || "")} />␊
           <div {...props} data-foo {...rest} className={"jsx-2886504620" + " " + (rest.className != null && rest.className || props.className != null && props.className || "")} />␊
           <div {...props} data-foo {...rest} className={"jsx-2886504620" + " " + (rest.className != null && rest.className || 'test')} />␊
+          <div {...{ id: 'foo' }} className={"jsx-2886504620"} />␊
+          <div {...{}} className={"jsx-2886504620" + " " + 'foo'} />␊
+          <div {...{ className: 'foo' }} className={"jsx-2886504620" + " " + "test"} />␊
+          <div className="test" {...{}} className={"jsx-2886504620" + " " + 'foo'} />␊
```
giuseppeg: nit: probably in this case we could remove the spread if it is empty

test/fixtures/attribute-generation-classname-rewriting.js
giuseppeg: looking good! We should add one last test case to make sure that it is handled correctly:

```jsx
<div className="test" {...{ className: props.className }} />
<div className="test" {...{ className: props.className }} {...bar } />
<div className="test" {...bar }  {...{ className: props.className }}  />
```

src/_utils.js
doniyor2109: Understand. If you want I can add function call support too?

src/_utils.js
```diff
@@ -28,12 +28,30 @@ export const addClassName = (path, jsxId) => {
...
+
+        const index = properties.findIndex(
+          property => property.key.name === 'className'
+        )
+
+        if (~index) {
+          className = attr.get('argument').get(`properties.${index}`)
+
+          // Remove jsx spread attribute if there is only className property
+          if (node.argument.properties.length === 1) {
```
giuseppeg: nit: here you can reference `properties` from L34

src/_utils.js
```diff
@@ -44,9 +62,7 @@ export const addClassName = (path, jsxId) => {
           attrNameDotClassName
         )
       )
-      continue
-    }
-    if (t.isJSXAttribute(attr) && node.name.name === 'className') {
+    } else if (t.isJSXAttribute(attr) && node.name.name === 'className') {
```
giuseppeg: why this change? I guess you can revert the continue

src/_utils.js
```diff
@@ -28,12 +28,30 @@ export const addClassName = (path, jsxId) => {
...
+        }
+        continue
+      }
+
       const name = node.argument.name
       const attrNameDotClassName = t.memberExpression(
-        t.isMemberExpression(node.argument)
-          ? node.argument
-          : t.identifier(name),
+        t.isIdentifier(node.argument) ? t.identifier(name) : node.argument,
```
giuseppeg: Sorry I had a second thought, we should probably not handle/support the function call and just let this fail. The reason being that we cannot make assumptions about the function and be sure that it is idempotent i.e. not sure if calling it multiple times  returns always the same thing. I would discourage inline calls like `{...foo()}`

giuseppeg:
@doniyor2109 amazing job! thank you

subtleGradient:
Yay!

PR #551, opened 2019-02-10
Fix source maps

### LINKED ISSUE #547: sourceMaps are broken
Related to #542

It seems that in 3.2.0 we broke source maps. This is probably because now we traverse manually. cc @ijjk

giuseppeg:
Fixes #547


src/_utils.js
```diff
@@ -591,7 +591,8 @@ export const processCss = (stylesInfo, options) => {
 
   if (useSourceMaps) {
     const generator = makeSourceMapGenerator(fileInfo.file)
-    const filename = fileInfo.sourceFileName
+    const filename = fileInfo.sourceFileName || fileInfo.filename
```
giuseppeg: The bug was here, these are not defined when not using the CLI I guess. Chances are that source maps still work when the plugin is used with babel loader or something.

ijjk: I started looking into this too and actually just opened a pr not seeing yours 🤦‍♂️

On my branch I used `file.opts.generatorOpts.sourceFileName` since it seems to match with the value we were expecting before. I closed it after seeing yours, but I just re-opened it so you can take a look.

src/_utils.js
```diff
@@ -563,18 +564,39 @@ export const processCss = (stylesInfo, options) => {
...
+    filename: file.opts.filename || file.filename
+  }
+
+  fileInfo.sourceFileName =
+    file.opts.sourceFileName ||
+    file.sourceFileName ||
+    // According to https://babeljs.io/docs/en/options#source-map-options
+    // filenameRelative = path.relative(file.opts.cwd, file.opts.filename)
+    // sourceFileName = path.basename(filenameRelative)
+    (fileInfo.filename &&
```
giuseppeg: @ijjk what do you think about this?

ijjk: This seems more stable than trying to follow where it's at in the options. Doesn't `path.basename` just return the last part of the path so the `path.relative` isn't necessary inside of the `path.basename` right?

giuseppeg: 👍 indeed.

PR #563, opened 2019-06-14
Remove singleton references to StylesheetRegistry.

### LINKED ISSUE #64: Instance-based API instead of global singleton
I think it would be nice if styled-jsx had a non-singleton interface.

The main use case I've found where a global singleton is problematic is related to [rendering within iframes](https://github.com/Khan/aphrodite/issues/130). In short, styles from components rendered within a iframe get rendered outside the iframe, preventing the styles from reaching inside the iframe.

The common pattern to solve this is to rely on React context to share instances, which allow the instance to be overridden within an iframe by overriding context. Then styles from components within a given iframe will be rendered within the iframe using its own instance.

This is somewhat uncommon use case, but creating a singleton interface from an instance is straightforward whereas going the reverse direction is much more difficult. I think making separate packages for the core and singleton interface would not add too much additional complexity or maintenance effort. I'd be happy to work on a PR, otherwise feel free to close this issue.

jamestalmage:
Apps must wrap with a StylesheetRegistryContext.Provider

Fixes #64 

There is definitely work left to be done before this can be merged, including documentation and probably some discussion of the API.

Most importantly, this introduces a number of breaking changes to the API:

  * `server.js`: `flushToReact()` and `flushToHtml()` both now require an instance of `StylesheetRegistry` as their first argument.
  * `style.js`: The `flush` function was removed. It was a static export.
  * `style.js`: The static class function `JSXStyle.dynamic(info)` was removed. It is not covered by tests, nor used anywhere in the project.
  * I introduced a whole bunch of helper functions in `server.js`. Most of them serve to wrap the API of [ReactDomServer](https://reactjs.org/docs/react-dom-server.html#overview). The wrapped functions return an object containing both the `result` of the called function (i.e. `renderToString`, `renderToStaticMarkup`, etc), and a new `StylesheetRegistry` that was passed to the render.
  * This makes use of `React.createContext`, so requires `React@16.3` or higher. If that is a problem, I can try to add support for the legacy API, but it's deprecated and will be removed.

TODO:

- [ ] Maintainers accept proposed API.
- [ ] Update documentation, especially SSR examples.
- [ ] PR to Next.js, which is broken by this change.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -1,8 +1,15 @@
 import React from 'react'
-import { flush } from './style'
+import StylesheetRegistryContext from './stylesheet-registry-context'
+import StylesheetRegistry from './stylesheet-registry'
 
-export default function flushToReact(options = {}) {
-  return flush().map(args => {
+function flush(styleSheetRegistry) {
+  const cssRules = styleSheetRegistry.cssRules()
+  styleSheetRegistry.flush()
+  return cssRules
+}
+
+export default function flushToReact(registry, options = {}) {
+  return flush(registry).map(args => {
     const id = args[0]
     const css = args[1]
     return React.createElement('style', {
@@ -17,8 +24,8 @@ export default function flushToReact(options = {}) {
   })
 }
 
-export function flushToHTML(options = {}) {
-  return flush().reduce((html, args) => {
+export function flushToHTML(registry, options = {}) {
+  return flush(registry).reduce((html, args) => {
     const id = args[0]
     const css = args[1]
     html += `<style id="__${id}"${
@@ -26,4 +33,55 @@ export function flushToHTML(options = {}) {
     }>${css}</style>`
     return html
   }, '')
+}
+
+export function wrapWithProvider(element, registry) {
+  return React.createElement(
+    StylesheetRegistryContext.Provider,
+    { value: registry },
+    element
+  )
+}
+
+export function callDomServerMethod(reactDomServer, method, element) {
+  const registry = new StylesheetRegistry()
+  const wrapped = wrapWithProvider(element, registry)
+  const result = reactDomServer[method](wrapped)
+  return { result, registry }
+}
+
+export function renderToString(server, element) {
+  return callDomServerMethod(server, 'renderToString', element)
+}
+
+export function renderToStaticMarkup(server, element) {
+  return callDomServerMethod(server, 'renderToStaticMarkup', element)
+}
+
+export function renderToNodeStream(server, element) {
+  return callDomServerMethod(server, 'renderToNodeStream', element)
+}
+
+export function renderToStaticNodeStream(server, element) {
+  return callDomServerMethod(server, 'renderToStaticNodeStream', element)
+}
+
+export function wrapServer(server) {
+  return {
+    renderToString(element) {
+      return renderToString(server, element)
+    },
+
+    renderToStaticMarkup(element) {
+      return renderToStaticMarkup(server, element)
+    },
+
+    renderToNodeStream(element) {
+      renderToNodeStream(server, element)
+    },
+
+    renderToStaticNodeStream(element) {
+      renderToStaticNodeStream(server, element)
+    }
+  }
 }
--- a/src/style.js
+++ b/src/style.js
@@ -1,7 +1,5 @@
 import { Component } from 'react'
-import StyleSheetRegistry from './stylesheet-registry'
-
-const styleSheetRegistry = new StyleSheetRegistry()
+import RegistryContext from './stylesheet-registry-context'
 
 export default class JSXStyle extends Component {
   constructor(props) {
@@ -9,7 +7,7 @@ export default class JSXStyle extends Component {
     this.prevProps = {}
   }
 
-  static dynamic(info) {
+  /* static dynamic(info) {
     return info
       .map(tagInfo => {
         const baseId = tagInfo[0]
@@ -17,7 +15,7 @@ export default class JSXStyle extends Component {
         return styleSheetRegistry.computeId(baseId, props)
       })
       .join(' ')
-  }
+  } */
 
   // probably faster than PureComponent (shallowEqual)
   shouldComponentUpdate(otherProps) {
@@ -30,7 +28,7 @@ export default class JSXStyle extends Component {
   }
 
   componentWillUnmount() {
-    styleSheetRegistry.remove(this.props)
+    this.context.remove(this.props)
   }
 
   render() {
@@ -39,10 +37,10 @@ export default class JSXStyle extends Component {
     if (this.shouldComponentUpdate(this.prevProps)) {
       // Updates
       if (this.prevProps.id) {
-        styleSheetRegistry.remove(this.prevProps)
+        this.context.remove(this.prevProps)
       }
 
-      styleSheetRegistry.add(this.props)
+      this.context.add(this.props)
       this.prevProps = this.props
     }
 
@@ -50,8 +48,4 @@ export default class JSXStyle extends Component {
   }
 }
 
-export function flush() {
-  const cssRules = styleSheetRegistry.cssRules()
-  styleSheetRegistry.flush()
-  return cssRules
-}
+JSXStyle.contextType = RegistryContext
--- a/src/stylesheet-registry-context.js
+++ b/src/stylesheet-registry-context.js
@@ -1,0 +1,27 @@
+import React from 'react'
+
+const defaultInstance = {}
+
+const methods = [
+  'add',
+  'remove',
+  'update',
+  'flush',
+  'cssRules',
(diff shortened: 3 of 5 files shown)


jamestalmage:
Spamming a few people who've expressed interest in this issue to gather feedback (sorry).

//cc @rauchg @rtsao @merrywhether @nkzawa @a-ignatov-parc

src/server.js
```diff
@@ -27,3 +34,54 @@ export function flushToHTML(options = {}) {
...
+      return renderToStaticMarkup(server, element)
+    },
+
+    renderToNodeStream(element) {
+      renderToNodeStream(server, element)
+    },
+
+    renderToStaticNodeStream(element) {
+      renderToStaticNodeStream(server, element)
+    }
```
jamestalmage: Not sure if I should be exporting all these helper methods. I'd like feedback on which style you guys like before adding tests/documentation.

timneutkens:
I don't think the provider wrapping should be provided (the helpers), it should be an implementation detail that users have to set up.

jamestalmage:
> I don't think the provider wrapping should be provided (the helpers), it should be an implementation detail that users have to set up.

Hmmm. It's a fair amount of code that I think will be boilerplate for most implementations. There is an argument to be made that I have gone overboard with the helpers. Not sure I'd agree with axing all of them.

timneutkens:
Could you write out the code for when a user sets it up instead of using the helpers? So that we can reference that when making the decision.

jamestalmage:
@timneutkens - went for a more aggressive refactor and ended up getting rid of all those helper functions.

`flushToHTML` is just `toHTML` on the registry now. And it doesn't flush anymore.

jamestalmage:
Do we even need a flush function anymore? Seems like it's pointless if we're instantiating a new registry on every render. Is the registry an expensive object? (doesn't appear to be)

jamestalmage:
Usage

```jsx
import React from 'react'
import ReactDOM from 'react-dom/server'
import StyleSheetRegistry, {RegistryProvider} from 'styled-jsx/registry'
import App from './app'

export default (req, res) => {
  const registry = new StyleSheetRegistry()
  const app = ReactDOM.renderToString(
    <RegistryProvider value={registry}>
      <App />
    </RegistryProvider>
  )
  const html = ReactDOM.renderToStaticMarkup(<html>
    <head>{ registry.toReact() }</head>
    <body>
      <div id="root" dangerouslySetInnerHTML={{__html: app}} />
    </body>
  </html>)
  res.end('<!doctype html>' + html)
}
```

src/stylesheet-registry-context.js
```diff
@@ -0,0 +1,35 @@
...
+      `StyleSheetRegistry.${methodName}: The style registry can no longer be referenced statically. You must wrap each render with a StyleSheetRegistryContext.Provider, and provide a unique registry for each render`
+    )
+  }
+})
+
+const StyleSheetRegistryContext = React.createContext(defaultInstance)
+
+export default StyleSheetRegistryContext
+
+export function wrapElementWithProvider(element, registry) {
```
giuseppeg: I think that this is unnecessary as users can do 

```jsx
<StyleSheetRegistryContext.Provider value={registry}>
  {children}
</StyleSheetRegistryContext.Provider>
```

src/style.js
```diff
@@ -1,24 +1,12 @@
...
-const styleSheetRegistry = new StyleSheetRegistry()
+import RegistryContext from './stylesheet-registry-context'
 
 export default class JSXStyle extends Component {
   constructor(props) {
     super(props)
     this.prevProps = {}
   }
 
-  static dynamic(info) {
```
giuseppeg: Please revert this. It computes dynamic classes at runtime. https://github.com/zeit/styled-jsx/blob/00fcb7bcad9c8af91167056f93c63858d24769ba/src/_utils.js#L347

src/_utils.js
```diff
@@ -142,7 +142,7 @@ export const validateExpressionVisitor = {
         (t.isThisExpression(node.object) &&
           (node.property.name === 'props' ||
             node.property.name === 'state' ||
-            node.property.name === 'context'))) ||
+            node.property.name === 'registry.js'))) ||
```
giuseppeg: I think you should revert this

src/style.js
```diff
@@ -1,24 +1,12 @@
 import { Component } from 'react'
-import StyleSheetRegistry from './stylesheet-registry'
-
-const styleSheetRegistry = new StyleSheetRegistry()
+import RegistryContext from './stylesheet-registry-context'
```
giuseppeg: To simplify things a bit, we could create the `Provider` and `Consumer` in this file and then export the `Provider`.

src/style.js
giuseppeg: In fact dynamic styles won't work with your solutions. Before we were relying on the singleton and using it in this static method. Now the method needs to be instance aware and therefore can't be static.

This is a problem because at build time (when we add `JSXStyle.dynamic` calls) we don't know anything about instances.

We'll need to think about a solution.

giuseppeg: probably you can move these outside of the class 

https://github.com/zeit/styled-jsx/blob/00fcb7bcad9c8af91167056f93c63858d24769ba/src/stylesheet-registry.js#L129-L195

make them IIFEs (the cache can be shared by all the instances of the registry I think) and export `computeId` so that `JSXStyle` can use it

src/stylesheet-registry-context.js
giuseppeg: In fact here is what I would do in `style.js`

```jsx
import StyleSheetRegistry from './stylesheet-registry'

const ctx = React.createContext()

  // ...
}

export function Provider({ registry, children }) {
  return (
    <ctx.Provider value={registry || new StyleSheetRegistry()}>
      {children}
    </ctx.Provider>
  )
}
```

src/stylesheet-registry.js
```diff
@@ -210,6 +211,33 @@ export default class StyleSheetRegistry {
...
+        key: `__${id}`,
+        nonce: options.nonce ? options.nonce : undefined,
+        dangerouslySetInnerHTML: {
+          __html: css
+        }
+      })
+    })
+  }
+
+  toHTML(options = {}) {
```
giuseppeg: I don't think that it these two method should be in the StylesheetRegistry as they are mostly for SSR - I'd revert the changes to `src/server.js`. We can save a tiny bit in client size bundle size. 

`src/style.js` could export the StylesheetRegistry singleton to flush the default when needed and these methods should accept an instance of StylesheetRegistry to flush.

By the way I think we need to `flush` instead of just callin `cssRules()`

jamestalmage: Why would you need to flush? Just discard the registry instance and let garbage collection do its thing.

src/_utils.js
jamestalmage: hmm. I believe IntelliJ did this when I renamed a file. Obviously not covered by tests.

src/stylesheet-registry-context.js
jamestalmage: I think you should move away from the singleton magic. It's just going to get more and more error prone as React continues to move towards async rendering / streaming rendering on the server.

Mutable singletons are bad. It shouldn't be your default.

src/stylesheet-registry.js
jamestalmage: As for moving those functions: https://martinfowler.com/bliki/TellDontAsk.html

src/style.js
jamestalmage: I wondered if `dynamic` was used by the Babel compiler, but I couldn't find a reference to it anywhere. This is another place where tests are obviously lacking.

src/stylesheet-registry-context.js
giuseppeg: If people need to do async rendering / streaming would still use multiple providers with their own instances of the registry. The ones who don't would need a default that is exactly the singleton. We can do that for them here and make their lives easier.

src/style.js
giuseppeg: yeah we don't have e2e tests because we need to set up some infra

src/stylesheet-registry.js
giuseppeg: If we keep the singleton we'd need to flush. 

I agree with you that the singleton is not cool but honestly we should aim to keep the DX good and just enable advanced stuff for the ones who are more experienced.

I'll let @timneutkens decide on this anyway

timneutkens:
Re: https://github.com/zeit/styled-jsx/pull/563#issuecomment-502329589

@jamestalmage this is exactly what I had in mind, awesome! Amazing work. Thank you so much 🙏

lifeiscontent:
@giuseppeg @timneutkens @jamestalmage any updates on this PR?

PR #572, opened 2019-08-06
Fix source maps pt. 2

### LINKED ISSUE #571: Source maps are still broken
The source maps are currently still broken after #547. After investigating the issue I found that a `sourceMaps` parameter in babel-external.js wasn't passed down to `processCss()` in _utils.js.
I created a PR with the fix.

bernhardc:
Fixes #571


giuseppeg:
thank you! I will release later today when I am at the computer

PR #581, opened 2019-09-10
CSSRulesList in browser environments is not actually an array.

emilsedgh:
Therefore it doesn't have a `map` function.
As a result, calling flush on browser environment fails.

This fixes the issue by converting it to an array using Array.from

### DIFF
--- a/src/lib/stylesheet.js
+++ b/src/lib/stylesheet.js
@@ -216,7 +216,7 @@ export default class StyleSheet {
     return this._tags.reduce((rules, tag) => {
       if (tag) {
         rules = rules.concat(
-          this.getSheetForTag(tag).cssRules.map(rule =>
+          Array.from(this.getSheetForTag(tag).cssRules).map(rule =>
             rule.cssText === this._deletedRulePlaceholder ? null : rule
           )
         )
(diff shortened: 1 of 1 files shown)


emilsedgh:
I know that it's not very typical to run styled-jsx/server in browser environments.
In my case, I'm creating a GUI editor for React (using [Grapesjs](https://github.com/artf/grapesjs)).

Nothing else prevents the code to work on a browser environment.

src/lib/stylesheet.js
```diff
@@ -216,7 +216,7 @@ export default class StyleSheet {
     return this._tags.reduce((rules, tag) => {
       if (tag) {
         rules = rules.concat(
-          this.getSheetForTag(tag).cssRules.map(rule =>
+          Array.from(this.getSheetForTag(tag).cssRules).map(rule =>
```
giuseppeg: Maybe go for the `Array.prototype.map.call(this.getSheetForTag(tag).cssRules, rule => {})` route so that when transpiling babel doesn't inline the polyfill for Array.from

emilsedgh: Of course. Patch upcoming.

giuseppeg:
awesome thank you for the contribution!

PR #591, opened 2019-11-07
Switch to loose mode

developit:
`styled-jsx` doesn't use any syntax that requires non-loose mode, and isn't using "spec" mode in the first place. Non-loose non-spec mode is verbose but doesn't adhere to the spec, so there's seldom a reason to use it.

This should reduce the size of styled-jsx considerably.

### DIFF
--- a/src/.babelrc
+++ b/src/.babelrc
@@ -1,4 +1,8 @@
 {
-  "presets": ["@babel/preset-env"],
-  "plugins": ["@babel/plugin-proposal-object-rest-spread"]
+  "presets": [
+    ["@babel/preset-env", { "loose": true }]
+  ],
+  "plugins": [
+    ["@babel/plugin-proposal-object-rest-spread", { "loose": true }
+  ]
 }
--- a/src/.babelrc
+++ b/src/.babelrc
@@ -3,6 +3,6 @@
     ["@babel/preset-env", { "loose": true }]
   ],
   "plugins": [
-    ["@babel/plugin-proposal-object-rest-spread", { "loose": true }
+    ["@babel/plugin-proposal-object-rest-spread", { "loose": true }]
   ]
 }
--- a/src/babel.js
+++ b/src/babel.js
@@ -159,7 +159,7 @@ export default function({ types: t }) {
             externalJsxId = t.templateLiteral(
               [
                 t.templateElement({ raw: 'jsx-', cooked: 'jsx-' }),
-                ...[...new Array(expressionsLength - 1)].map(() =>
+                ...[...new Array(expressionsLength - 1).fill(null)].map(() =>
                   t.templateElement({ raw: ' jsx-', cooked: ' jsx-' })
                 ),
                 t.templateElement({ raw: '', cooked: '' }, true)


src/.babelrc
```diff
@@ -1,4 +1,8 @@
 {
-  "presets": ["@babel/preset-env"],
-  "plugins": ["@babel/plugin-proposal-object-rest-spread"]
+  "presets": [
+    ["@babel/preset-env", { "loose": true }]
+  ],
+  "plugins": [
+    ["@babel/plugin-proposal-object-rest-spread", { "loose": true }
```
giuseppeg:
```suggestion
    ["@babel/plugin-proposal-object-rest-spread", { "loose": true }]
```

developit:
Just triple-checked, it's most definitely loose mode compatible.

I think loose mode gets a bad reputation without much reason. The cases where it's less compatible than non-loose mode are extremely rare and easily avoided. Babel is actually looking into removing the middle "not strict but not loose" option since it's bit of a weird grey area. In general, loose mode drops support for edge cases where static analysis fails - detection and handling of Iterables, funky class inheritance chains (super constructors that throw or return exocit objects), etc. Most libraries operate in an environment where this is never needed, and the payoff for telling Babel this up-front is high.

Here's an analysis of each file and its loose mode compatibility:

#### `src/stylesheet-registry.js`

- classes are loose mode compatible since they have no inheritance and only methods/properties. It could technically be compiled to a straight prototype assignment, but that's for another time
- template literals remain unchanged, no tagged templates in use
- no rest/spread is used, so no change there
- default parameter value for the constructor is the easy case since it's single-arity - loose mode compiles this to `x = x!==undefined ? x : {}` rather than the current `arguments.length` check.

> Loose mode reduces the minified+gzipped size of this module by 15%

#### `src/style.js`:
- the react component class is also loose mode compatible. It inherits from `react.Component` which we know never has a return value or throws.
- the `static dynamic()` method will just get compiled to `JSXStyle.dynamic = function(){...}`, which is unchanged from its current form.

> Loose mode reduces the minified + gzipped size of this module by 50%.

### `src/lib/style-transform.js`:

- The `...args` usage is loose mode compatible. It's actually poorly optimized in all Babel modes, and I am working with the Babel team to address this. We should reconvene once that work is released.
- Template Literal usage compiles to string concatenation instead of `String.concat()`, which is smaller while retaining identical functionality for the use-case.

> Loose mode only reduces the size of this module minimally, around 5-10%. Expect much better reduction from Babel soon.

### `src/lib/stylesheet.js`:

- there's very little ES6 in this file, it's actually Bublé-compatible and certainly loose mode compatible.
- the `length` getter makes loose mode output a bit more boilerplate than the other modules do, but it's still compatible.

> Size reduction of 5%. There is some low-hanging fruit here for optimization (`isString`/`isProd`/`invariant`)

In total I expect loose mode to reduce the size of the styled-jsx runtime by around 20%. I'd say that's worth it!

giuseppeg:
@developit thank you for the detailed reply! 

> the length getter makes loose mode output a bit more boilerplate than the other modules do, but it's still compatible.

Probably in a follow up we could rewrite this (I think it is used only in tests atm). 

Can you run the tests locally and see what's going on?

giuseppeg:
Probably Babel is transpiling everything using babelrc and loose mode breaks the plugin. I am not even sure if it makes sense to transpile the babel plugin and friends. Maybe we should build only the browser code or otherwise have two separate scripts.

developit:
@giuseppeg I'll clone locally and see what's up. It might make sense to move this to microbundle at some point for the size improvements and property mangling - I think I may even have a local fork that I converted...

developit:
Found it, there was a `[...new Array()]` bit that needed `.fill()`.

giuseppeg:
awesome thank you @developit!

PR #592, opened 2019-11-07
Escape filename in generated source map code.

emielta:
Currently on windows `sourceURL`s for source maps are broken and if the path includes a file or folder that starts with "u" or "x" followed by an incorrect escape sequence the produced code fails to compile.

This replaces all instances of "\" inside the string with "\\" resulting in properly escaped backslashes.

Issue on next.js repo: https://github.com/zeit/next.js/issues/9065


giuseppeg:
@emielta thank you!

Issue #618, opened 2020-01-27
Upgrade to babel v7


ocavue:
#### Do you want to request a *feature* or report a *bug*?

bug 

#### What is the current behavior?

`styled-jsx` v3.2.4 requires some very old (2 years+) babel dependencies: `babel-plugin-syntax-jsx@6.18.0` and `babel-types@6.26.0`. Maybe it's time to upgrade to babel v7.

BTW, since babel v7, babel changed it's package names. `babel-types` => `@babel/types` for example.

#### If the current behavior is a bug, please provide the steps to reproduce and possibly a minimal demo or testcase in the form of a Next.js app, CodeSandbox URL or similar

I found this babel version issue because of a polyfill issue of ES6 `Map` from my Next.js project. In short, the polyfill of `Map` from babel v6 make `obj instanceof Map` always return `false`.

I think the request of "upgrade babel to v7" doesn't really need a special reason. However, I can describe the `Map` issue above with more details if your guys think it's necessary.

#### What is the expected behavior?

`styled-jsx` use babel v7 as dependency.

#### Environment (include versions)

* OS: N/A
* Browser: N/A
* styled-jsx (version): v3.2.4

#### Did this work in previous versions?

No

giuseppeg:
see description https://github.com/zeit/styled-jsx/pull/621

PR #621, opened 2020-02-14
Upgrade to @babel/types to remove transitive babel-runtime depenency

tkalliom:
This change allows usage of `styled-jsx` without pulling in the old `babel-runtime`, but still retains compatibility with Babel 6.

Relates to #618 (which, however, IMHO is not really a bug, as Babel 6 compatibility is apparently a conscious decision)


giuseppeg:
thank you @tkalliom !

PR #629, opened 2020-04-08
Adds support for styleModule

arcanis:
This PR adds support for the `styleModule` configuration, a way to change the (currently hardcoded) module that gets imported during the styled-jsx transpilation.

Part 1 of https://github.com/zeit/next.js/issues/10149

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -13,6 +13,7 @@ Code and docs are for v3 which we highly recommend you to try. Looking for style
 - [Configuration options](#configuration-options)
   * [`optimizeForSpeed`](#optimizeforspeed)
   * [`sourceMaps`](#sourcemaps)
+  * [`styleModule`](#stylemodule)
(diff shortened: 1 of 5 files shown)


giuseppeg:
@arcanis thanks for the PR! Looking at the issue I wonder if this could be fixed in Next.js by rewriting the import declaration source with a Babel plugin so that it points to the right module or with a webpack alias?

arcanis:
I think idiomatically the 'right' path should be configurable by the transforms themselves - this is for example what `transform-runtime` does (kinda) with [`absoluteRuntime`](https://babeljs.io/docs/en/babel-plugin-transform-runtime#absoluteruntime) which instructs it to use the absolute path rather than the regular package name. In our case I think it's better for portability to avoid absolute paths and use the Next.js package name instead.

giuseppeg:
Thank you @arcanis and sorry for the delay! I'll release a minor version

PR #647, opened 2020-05-27
Add support for custom HTML tags

ottobonn:
This PR is my follow-up to the discussion in https://github.com/zeit/next.js/discussions/13376. It adds support for custom HTML tags (whose names are lower case and contain hyphens) to styled-jsx.

Here's an example of such a tag in React:

```js
export default function Test() {
  return (
    <my-test class="root">
      <style jsx>{`
        .root {
          margin: 0;
          width: 100%;
          height: 100%;
        }
      </style>
    </my-test>
  );
}
```

Before this PR, the output of styled-jsx would be:

```html
<my-test class="root" className="jsx-12345">
```

This output is not quite right, because React passes through the raw `class` attribute of the element for custom HTML tags, and styled-jsx is attaching additional class names to `className`. 

This PR adds code to detect custom element names and to use `class` instead of `className` for them.


ottobonn:
Looks like the CI build failed due to the Zeit/Vercel rename...

heff:
Thanks for figuring this out @ottobonn! Running into this issue with people trying to use styled-jsx with [media-chrome](https://github.com/muxinc/media-chrome) custom elements. Is there something I can do to help get this over the line?

PR #651, opened 2020-06-30
Adding ability to select components

artalar:
> Inspired by [reshadow](https://reshadow.dev)

Now you may select components in styles if that components accepts `className` property:

```jsx
import React from 'react'
import Link from 'some-library'
export default () => (
  <div>
    <Link>About</Link>
    
    <style jsx>{`
      .Link {
        text-decoration: none;
      }
    `}</style>
  </div>
)
```

The next question is how to apply styles conditionally...


artalar:
@giuseppeg  what do you think?

huozhi:
Thanks for the PR! This approach might birng some confusion to users since it doesn't follow the css selector rules. Currently it's not easy to support use component name as selector since the code minification could change it during optimization phase. Would like to keep current as-is 🙏

PR #676, opened 2020-11-13
Upgrade Babel deps and fix pure computation

Timer:
This upgrades all of our dev dependencies to the newest versions and replaces our custom pure evaluator with Babel's built-in scope evaluator.


timneutkens (approved):
:100:

ijjk (approved):

src/_utils.js
```diff
@@ -294,7 +225,21 @@ export const getJSXStyleInfo = (expr, scope) => {
 
   const { quasis, expressions } = node
   const hash = hashString(expr.getSource().slice(1, -1))
-  const dynamic = scope ? isDynamic(expr, scope) : false
+  let dynamic = Boolean(scope)
+  if (dynamic) {
+    try {
+      const val = expr.evaluate()
```
giuseppeg, line 231: I think that we could do the same to replace `validateExternalExpressions`

giuseppeg (commented):
💯

PR #677, opened 2020-11-14
fix(external): execute transforms before JSX transform

Timer:
The external visitors need to run before other transforms, just like the main visitors.

This is necessary because the external transform generates JSX, which needs to be processed by the JSX transform, which requires it be present before the "main" Program execution.

---

Fixes https://github.com/vercel/next.js/issues/18134


Timer:
cc @timneutkens @ijjk

ijjk (approved):

gtolarc:
It seems that the version with this commit has not yet been released on public npm.

PR #684, opened 2020-11-19
fix: make styled-jsx/babel plugin respect the source type

jaydenseric:
Fixes https://github.com/vercel/styled-jsx/issues/680 .

[`@babel/helper-module-imports`](https://babeljs.io/docs/en/babel-helper-module-imports) should be used when injecting imports, as it will use either `import` or `require` syntax depending if the Babel `sourceType` for the file is `module` (ESM) or `script` (CJS).


package.json
```diff
@@ -52,6 +52,7 @@
     }
   },
   "dependencies": {
+    "@babel/helper-module-imports": "7.12.5",
```
jaydenseric, line 55: Pinning an exact version (without a `^`) is 100% an anti-pattern for a published package, but I know from past discussions your mind can't be changed so I made it consistent with the other dependencies.

src/_utils.js
```diff
@@ -621,14 +622,12 @@ export const booleanOption = opts => {
...
+    state.file.path,
     typeof state.opts.styleModule === 'string'
       ? state.opts.styleModule
-      : 'styled-jsx/style'
-
-  return t.importDeclaration(
-    [t.importDefaultSpecifier(t.identifier(STYLE_COMPONENT))],
-    t.stringLiteral(styleModule)
+      : 'styled-jsx/style',
+    { nameHint: STYLE_COMPONENT}
```
jaydenseric, line 630: I'm not 100% sure what will happen if there is an existing variable in scope that has the same name as `STYLE_COMPONENT` (`_JSXStyle`), but I think the `addDefault` helper will modify the `nameHint` to avoid a collision. Ideally, no `nameHint` would be used and the automatic name of the injected default import would be used throughout the code instead of a hardcoded constant. This way a user could have a `_JSXStyle` variable name in the same scope, no problem.

package.json
jaydenseric: Note that by pinning an exact version, if a user already has a newer patch than `7.12.5` in their `node_modules`, installing `styled-jsx` will downgrade it, or create duplicate installations of `@babel/helper-module-imports` at different versions within `node_modules`. Nasty.

Timer:
Happy to land this if you update it!

src/_utils.js
Timer: All code branches that call this already check if `STYLE_COMPONENT` is in scope, so this should "just work".

giuseppeg: There is a remote chance that `_JSXStyle` is in the source code i.e. the user defined a binding with that name which will conflict with `STYLE_COMPONENT`. 

If you want to fix this, you should probably [generate a unique identifier](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#generating-a-uid) from `STYLE_COMPONENT`, put it in state and pass it to helpers if necessary but I probably I wouldn't bother in this PR – it's been like that for 5-ish years

jaydenseric:
I edited the PR title to conform with the failing checks, but the `build` check has been stuck on "Expected — Waiting for status to be reported" for 17 minutes, so I think something is wrong with the Actions.

@Timer 

> Happy to land this if you update it!

Where you just referring to the PR title? This PR is good-to-go.

Timer:
I swore I saw a merge conflict, but I guess not.

Timer (approved):

PR #699, opened 2021-02-16
fix: Typing module location and invalid export

### LINKED ISSUE #698: Bad TypeScript types
#### Do you want to request a *feature* or report a *bug*?

Bug/issue.

#### What is the current behavior?

The current types list typings in package.json which claim that `styled-jsx` exports `css`... which isn't true. You actually need to import `styled-jsx/css`.

This also requires there to be a root level `css.d.ts` file (at the same level as `css.js` in the root of the package, not inside `typings/`).

#### Environment (include versions)

* styled-jsx (version): 3.4.2

jrandiny:
Fix #698


Timer (approved):

travigd:
Thank you, friend!

PR #703, opened 2021-02-19
Allow providing custom StyleSheetRegistry (fix async render issues)

### LINKED ISSUE #64: Instance-based API instead of global singleton

travigd:
Should fix #64 (while trying to keep existing workflows working and changing as few lines of code as possible).

# Changes
* Introduce `StyleSheetRegistryContext` (whose default value is `globalStyleSheetRegistry` -- means should work without changes for existing users who don't care about async rendering)
* Bump React requirement to 16.3 to get new context. (we **could** get around this if people feel strongly, but... React 16.3 should be a decent minimum in my _oh so humble_ opinion)

---

This is still somewhat of a WIP, but I think it's at the point that it could be review.

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -1,7 +1,8 @@
-import { Component } from 'react'
-import StyleSheetRegistry from './stylesheet-registry'
-
-const styleSheetRegistry = new StyleSheetRegistry()
+import React, { Component } from 'react'
+import {
+  globalStyleSheetRegistry,
+  StyleSheetRegistryContext
+} from './stylesheet-registry'
 
 export default class JSXStyle extends Component {
   constructor(props) {
@@ -9,12 +10,20 @@ export default class JSXStyle extends Component {
     this.prevProps = {}
   }
 
+  get styleSheetRegistry() {
+    return this.context
+  }
+
   static dynamic(info) {
     return info
       .map(tagInfo => {
         const baseId = tagInfo[0]
         const props = tagInfo[1]
-        return styleSheetRegistry.computeId(baseId, props)
+        // It's fine te reference globalStyleSheetRegistry here since all
+        // `computeId` does is calculate the classname (which is a pure function
+        // of baseId and props). It's only an instance method so that it can
+        // memoize the results.
+        return globalStyleSheetRegistry.computeId(baseId, props)
       })
       .join(' ')
   }
@@ -30,7 +39,7 @@ export default class JSXStyle extends Component {
   }
 
   componentWillUnmount() {
-    styleSheetRegistry.remove(this.props)
+    this.styleSheetRegistry.remove(this.props)
   }
 
   render() {
@@ -39,10 +48,10 @@ export default class JSXStyle extends Component {
     if (this.shouldComponentUpdate(this.prevProps)) {
       // Updates
       if (this.prevProps.id) {
-        styleSheetRegistry.remove(this.prevProps)
+        this.styleSheetRegistry.remove(this.prevProps)
       }
 
-      styleSheetRegistry.add(this.props)
+      this.styleSheetRegistry.add(this.props)
       this.prevProps = this.props
     }
 
@@ -50,7 +59,9 @@ export default class JSXStyle extends Component {
   }
 }
 
-export function flush() {
+JSXStyle.contextType = StyleSheetRegistryContext
+
+export function flush(styleSheetRegistry = globalStyleSheetRegistry) {
   const cssRules = styleSheetRegistry.cssRules()
   styleSheetRegistry.flush()
   return cssRules
--- a/src/stylesheet-registry.js
+++ b/src/stylesheet-registry.js
@@ -1,5 +1,6 @@
 import hashString from 'string-hash'
 import DefaultStyleSheet from './lib/stylesheet'
+import React from 'react'
 
 const sanitize = rule => rule.replace(/\/style/gi, '\\/style')
 export default class StyleSheetRegistry {
@@ -211,6 +212,11 @@ export default class StyleSheetRegistry {
     }, {})
   }
 }
+
+export const globalStyleSheetRegistry = new StyleSheetRegistry()
+export const StyleSheetRegistryContext = React.createContext(
+  globalStyleSheetRegistry
+)
 
 function invariant(condition, message) {
   if (!condition) {
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "xo": "0.24.0"
   },
   "peerDependencies": {
-    "react": "15.x.x || 16.x.x || 17.x.x"
+    "react": "16.3 || 17"
   },
   "keywords": [
     "babel-plugin-macros",
--- a/server.d.ts
+++ b/server.d.ts
@@ -2,10 +2,22 @@
 
 import { ReactElement } from 'react'
 
-declare function flushToHTML(opts?: { nonce?: string }): string
-declare function flushToReact<T>(opts?: {
-  nonce?: string
-}): Array<ReactElement<T>>
+export declare class StyleSheetRegistry {
+  constructor()
 
-export { flushToHTML }
+  // These are marked private just so that we don't have to deal with specifying
+  // their return types (but they need to be here so that users don't just pass
+  // arbitrary objects in its place).
+  private cssRules()
+  private flush()
+}
+
+export interface FlushOpts {
+  nonce?: string
+  registry?: StyleSheetRegistry
+}
+
+export declare function flushToHTML(opts?: FlushOpts): string
+
+declare function flushToReact<T>(opts?: FlushOpts): Array<ReactElement<T>>
 export default flushToReact
--- a/src/server.js
+++ b/src/server.js
@@ -1,8 +1,18 @@
 import React from 'react'
 import { flush } from './style'
 
+import StyleSheetRegistry, {
+  StyleSheetRegistryContext,
+  globalStyleSheetRegistry
+} from './stylesheet-registry'
+export {
+  StyleSheetRegistry,
+  StyleSheetRegistryContext,
+  globalStyleSheetRegistry
+}
+
 export default function flushToReact(options = {}) {
-  return flush().map(args => {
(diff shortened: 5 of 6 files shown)


travigd:
This does seem to work in my NextJS codebase.

```tsx
// _document.tsx (abridged)

class AppDocument extends Document {
  static async getInitialProps(ctx: DocumentContext) {
    // create a per-request registry
    const styledJsxRegistry = new StyleSheetRegistry();

    // override renderPage (as suggested in NextJS documentation)
    const originalRenderPage = ctx.renderPage;
    ctx.renderPage = (opts) => {
      return originalRenderPage({
        ...opts,
        enhanceApp: (App) => (props) => (
          <StyleSheetRegistryContext.Provider value={styledJsxRegistry}>
            <App {...props} />
          </StyleSheetRegistryContext.Provider>
        ),
      });
    };

    const initialProps = await Document.getInitialProps(ctx);
    const styledJsxStyles = flushToReact({ registry: styledJsxRegistry });

    return {
      ...initialProps,
      styles: (
        <>
          {initialProps.styles}
          {styledJsxStyles}
        </>
      ),
    };
  }

  ...
}
```

src/style.js
```diff
@@ -1,20 +1,29 @@
...
     return info
       .map(tagInfo => {
         const baseId = tagInfo[0]
         const props = tagInfo[1]
-        return styleSheetRegistry.computeId(baseId, props)
+        // It's fine te reference globalStyleSheetRegistry here since all
+        // `computeId` does is calculate the classname (which is a pure function
+        // of baseId and props). It's only an instance method so that it can
+        // memoize the results.
+        return globalStyleSheetRegistry.computeId(baseId, props)
```
giuseppeg, line 26: Correct! Can you instead move methods outside of the StyleSheetRegistry class? This way we don't need to use the `globalStyleSheetRegistry` at all.

eg:

```js
export const computeId = (function createComputeId() {
  const cache = {}
  return function(baseId, props) {
    if (!props) {
      return `jsx-${baseId}`
    }

    const propsToString = String(props)
    const key = baseId + propsToString
    // return `jsx-${hashString(`${baseId}-${propsToString}`)}`
    if (!cache[key]) {
      cache[key] = `jsx-${hashString(`${baseId}-${propsToString}`)}`
    }

    return cache[key]
  }
}());
```

src/stylesheet-registry.js
```diff
@@ -212,6 +213,11 @@ export default class StyleSheetRegistry {
   }
 }
 
+export const globalStyleSheetRegistry = new StyleSheetRegistry()
```
giuseppeg, line 216: Unfortunately when the constructor of `StyleSheetRegistry` creates a new StyleSheet and injects it into the page. In the case when you don't use the global registry this would be useless.

Perhaps we can create a default and shared `StyleSheetRegistry` in `style.js` only when `this.context` is undefined?

giuseppeg (commented):
Hi @travigd thanks for the PR!! So far so good! Can you add some tests? @ me when the PR is ready. 

A few comments:

src/stylesheet-registry.js
travigd: Two things:

**First thing:**
The predominant use case for this will be server rendering, where the sheet **isn't** injected (since it's on the server). When client rendering, we can just use the global instance.

**Second thing:**
Are you sure about that assessment? Either way, the `StyleSheet` gets injected when `StyleSheetRegistry` is constructed. If that happens only during SSR, it's a no-op, but if you're using `StyleSheetRegistryContext` as part of CSR, you'll still presumably construct a new `StyleSheetRegistry` and it will be injected into the DOM.

travigd:
@giuseppeg I'm happy with this PR.

If we **really** want to preserve compat with React <16.3, we can probably hack it in (with conditional `if (typeof React.createContext !== undefined)` statements). But React 16.3 was released almost three years ago.

src/stylesheet-registry.js
giuseppeg: > Either way, the StyleSheet gets injected when StyleSheetRegistry is constructed.

Correct

> but if you're using StyleSheetRegistryContext as part of CSR

If you used multiple `StyleSheetRegistry` each using a dedicated StyleSheet, you'll have the same amount of StyleSheets on the client (due to hydration the same code that ran on the server will run on the client).

What I meant with my previous comment is that if you don't use the `globalStyleSheetRegistry` instance, on the client you'll have one extra (unused) <style> element. 

Probably not a big deal and the following is not worth and ugly, but the above could be avoided if in `style.js` you do the following:

```jsx
let globalStyleSheetRegistry = null
  // ... 
  get styleSheetRegistry() {
      if (this.context instanceof StyleSheetRegistry) { 
          return this.context
      }
      
       if (!globalStyleSheetRegistry) {
              globalStyleSheetRegistry = new StyleSheetRegistry()
       }
       return globalStyleSheetRegistry
   }
}
```

with 

```js
export const StyleSheetRegistryContext = React.createContext(null)
```

readme.md
```diff
@@ -391,11 +390,68 @@ It's **paramount** that you use one of these two functions so that
...
+### Asynchronous rendering
+
+`styled-jsx` defaults to using a single, global store for styles that are added
+while rendering. While this works for simple sites (including statically
+rendered pages), it can cause issues if using asynchronous rendering.
+
+To resolve these issues, you can wrap your app in a
+`StyleSheetRegistryContext.Provider` that is unique to the request.
+
+Using NextJS, your `_document.jsx` should look like this:
```
giuseppeg, line 402: Maybe this example can go in a <detail> element so it doesn't take so much space?

src/style.js
```diff
@@ -1,20 +1,22 @@
...
+    return this.context
+  }
+
   static dynamic(info) {
     return info
       .map(tagInfo => {
         const baseId = tagInfo[0]
         const props = tagInfo[1]
-        return styleSheetRegistry.computeId(baseId, props)
+        return computeId(baseId, props)
```
giuseppeg, line 19: Is there any reason for not passing a `cache` here? Probably it is ok to export a version with a built in cache (the iife) I wrote in the other comment.

readme.md
travigd: What do you mean by _in a element_? Like a `<spoiler />` tag?

src/style.js
travigd: I can do that. My only worry is just that it might become a bit of a memory leak since you have a cache entry per unique set of props (so if you have a long running app that uses different interpolations, they'd never get cleared).

What do you think?

src/stylesheet-registry.js
travigd: Oh, I see what you mean.

Personally, I think having the global registry just always present is fine even if it's empty (especially since I suspect the predominant use case will only be wrapping it for a SSR then using the global registry for everything on the client). My NextJS example that I added doesn't suffer from this for exactly that reason (the _document code only runs on the server).

If you'd really like me to change something, let me know I can do it. But like I said, I think the current behavior in this PR is fine.

travigd:
Hey @giuseppeg , just bumping this. No huge rush but would love to get this merged and released. 🎉

src/style.js
giuseppeg: Oh well it has been like this for ever and nobody has reported any issue of this kind. If necessary we can replace the object with a WeakMap any time though (later)

readme.md
giuseppeg: yes I meant <details><summary> ... github cut off `<details>` because I forgot to put it in back ticks :)

giuseppeg:
Hi @travigd sorry for making you wait – I haven't had time to check the PR again. But it looks good, there are just a couple of unresolved comments (especially the one about the cache) that might be worth fixing.

@timneutkens can you take a look at this PR when you have time? I just want to make sure that these changes won't conflict with future plans of yours (assuming styled-jsx will be updated to support concurrent mode or server components are out).

travigd:
Yep! Thank you! I did respond to the caching thing as being a potential memory leak if we use a global cache that is never flushed during SSR. :)

travigd:
ping @timneutkens

travigd:
@timneutkens @giuseppeg Any way we can move forward with this? :)

timneutkens:
Overall this looks good, however I'm not convinced we should provide backwards compat with the previous approach, it'll likely be a lot simpler to release a new major version that does not support the global registry which is an anti-pattern nowadays anyway 👍

travigd:
If I get a say, I'd suggest releasing this as a minor then re-write the code base to include a few breaking changes (like this) and also adopt typescript and maybe reduce reliance on snapshot testing while we're at it.

timneutkens:
Overall there's no benefit in releasing a minor though as styled-jsx is mostly used with Next.js.

src/style.js
giuseppeg: I'd say let's fix this (bring back the cache)

travigd:
Maybe I'm confused, but... I was able to install this version onto my NextJS project and it works just fine.

giuseppeg:
This issue https://github.com/vercel/styled-jsx/issues/714 made me realize that if we want to support multiple registries we'd need to find a way to hydrate SSR styles.

Currently the first instance of the stylesheet registry will handle those styles 

https://github.com/vercel/styled-jsx/blob/27418790cc3031113d031e2742ecdaab197e18a3/src/stylesheet-registry.js#L202-L205

PR #720, opened 2021-07-09
feat: use react hooks to manage styles injection

huozhi:
StyleSheetRegistry is an external state outside of react, to keep sync with react internal state, we'd better use reliable lifecycle to count the change instead of use cached class properties and `render` function to determine changes.

### Lifecycle
first render -> add style
update -> remove then add
unmount -> remove

### Problem
the issue is the react internal state might be different with the cached class properties `this.prevProps`, react might drop some render due to prioritise change, leading to **tearing**.

### Notice
This PR only solved the effects issue when using the react under concurrent mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "xo": "0.24.0"
   },
   "peerDependencies": {
-    "react": "15.x.x || 16.x.x || 17.x.x"
+    "react": "17.x.x || 18.x.x"
   },
   "keywords": [
     "babel-plugin-macros",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "xo": "0.24.0"
   },
   "peerDependencies": {
-    "react": "17.x.x || 18.x.x"
+    "react": ">= 16.8.0 || 17.x.x || 18.x.x"
   },
   "keywords": [
     "babel-plugin-macros",
--- a/src/style.js
+++ b/src/style.js
@@ -1,53 +1,30 @@
-import { Component } from 'react'
+import { useLayoutEffect } from 'react'
 import StyleSheetRegistry from './stylesheet-registry'
 
 const styleSheetRegistry = new StyleSheetRegistry()
 
-export default class JSXStyle extends Component {
-  constructor(props) {
-    super(props)
-    this.prevProps = {}
-  }
-
-  static dynamic(info) {
-    return info
-      .map(tagInfo => {
-        const baseId = tagInfo[0]
-        const props = tagInfo[1]
-        return styleSheetRegistry.computeId(baseId, props)
-      })
-      .join(' ')
-  }
-
-  // probably faster than PureComponent (shallowEqual)
-  shouldComponentUpdate(otherProps) {
-    return (
-      this.props.id !== otherProps.id ||
-      // We do this check because `dynamic` is an array of strings or undefined.
-      // These are the computed values for dynamic styles.
-      String(this.props.dynamic) !== String(otherProps.dynamic)
-    )
-  }
-
-  componentWillUnmount() {
-    styleSheetRegistry.remove(this.props)
-  }
-
-  render() {
-    // This is a workaround to make the side effect async safe in the "render" phase.
-    // See https://github.com/zeit/styled-jsx/pull/484
-    if (this.shouldComponentUpdate(this.prevProps)) {
-      // Updates
-      if (this.prevProps.id) {
-        styleSheetRegistry.remove(this.prevProps)
-      }
-
-      styleSheetRegistry.add(this.props)
-      this.prevProps = this.props
-    }
-
+export default function JSXStyle(props) {
+  if (typeof window === 'undefined') {
+    styleSheetRegistry.add(props)
     return null
   }
+  useLayoutEffect(() => {
+    styleSheetRegistry.add(props)
+    return () => {
+      styleSheetRegistry.remove(props)
+    }
+  }, [props.id, props.children, String(props.dynamic)])
+  return null
+}
+
+JSXStyle.dynamic = info => {
+  return info
+    .map(tagInfo => {
+      const baseId = tagInfo[0]
+      const props = tagInfo[1]
+      return styleSheetRegistry.computeId(baseId, props)
+    })
+    .join(' ')
 }
 
 export function flush() {
--- a/src/style.js
+++ b/src/style.js
@@ -13,7 +13,8 @@ export default function JSXStyle(props) {
     return () => {
       styleSheetRegistry.remove(props)
     }
-  }, [props.id, props.children, String(props.dynamic)])
+    // props.children can be string[], will be striped since id is identical
+  }, [props.id, String(props.dynamic)])
   return null
 }
 


shuding:
Is this caused by strict effects or is it something else? Could you link more references in the description? Would appreciate it!

> componentWillUnmount is keeping called from react flushSyncCallbacks in react 18

package.json
```diff
@@ -91,7 +91,7 @@
     "xo": "0.24.0"
   },
   "peerDependencies": {
-    "react": "15.x.x || 16.x.x || 17.x.x"
+    "react": "16.x.x || 17.x.x"
```
shuding, line 94: Do we need to drop React 15?

huozhi: introduced new lifecycle `getDerivedStateFromProps` from react 16

Timer:
If you end up dropping an old react version, make sure you add the following as a new line to the commit message when squash & merged:

```
BREAKING CHANGE: This change drops support for React 15.
```

Adding that line to the commit message will ensure this releases as a new major!

src/style.js
```diff
@@ -1,53 +1,30 @@
...
-    styleSheetRegistry.remove(this.props)
+export default function JSXStyle(props) {
+  if (typeof window === 'undefined') {
+    styleSheetRegistry.add(props)
+    return null
   }
-
-  render() {
-    // This is a workaround to make the side effect async safe in the "render" phase.
-    // See https://github.com/zeit/styled-jsx/pull/484
```
giuseppeg, line 38: Please read the comment below and see if you can verify that everything works fine i.e. triggering layout in a parent component doesn't run transitions. Before this was due to the fact that styles are appended after the DOM is updated. 

@sebmarkbage explained what's going on here https://github.com/vercel/styled-jsx/pull/484#issuecomment-584474687 This is by the way the reason why I was adding and removing styles during rendering - I needed them to be in the page before the related component was committed/updated

src/style.js
```diff
@@ -1,53 +1,30 @@
...
-      }
-
-      styleSheetRegistry.add(this.props)
-      this.prevProps = this.props
+  useLayoutEffect(() => {
+    styleSheetRegistry.add(props)
+    return () => {
+      styleSheetRegistry.remove(props)
     }
+  }, [props.id, String(props.dynamic)])
```
giuseppeg, line 16: id` is an hash of `children` so it might be fine to omit children but for correctness `children` should be in the deps array because it is a dependency too.

src/style.js
huozhi: `useLayoutEffect` is safer than `useEffect` here since `useLayoutEffect` inject the styles to DOM synchronously after DOM mutations which kind avoid the repaint happening very soon by browser.

We tried the jsfiddle example provided in that PR to test with react 17. If style injection happens in `useEffect` then it repaint but it doesn't repaint with useLayoutEffects.

Also found that emotion and styled-components both use `useLayoutEffects` to manage style injection. So I think this so be a fine solution so far? Before react releases any new hook for this case.

giuseppeg: Transitions are still run when you trigger layout https://codesandbox.io/s/react-17-repaint-demo-forked-c716t?file=/src/App.js 

I'll leave it to you to decide whether this is ok but if I remember correctly I had to come up with that hack (inject while rendering) because of a bug report from Vercel folks.

shuding: As Sebastian said in that comment, it's missing a good solution and it feels OK for styled-jsx at the moment, as it's more like an uncovered case of React. We should either [look forward to an official React API for this](https://github.com/reactwg/react-18/discussions/69#discussioncomment-995228), or avoid immediate layout triggering in the application code.

I'll approve this PR since there will be a much bigger issue with the current implementation when enabling concurrent rendering in React 18.

src/style.js
```diff
@@ -1,53 +1,30 @@
...
-      String(this.props.dynamic) !== String(otherProps.dynamic)
-    )
-  }
-
-  componentWillUnmount() {
-    styleSheetRegistry.remove(this.props)
+export default function JSXStyle(props) {
+  if (typeof window === 'undefined') {
+    styleSheetRegistry.add(props)
+    return null
```
shuding, line 9: If we SSR the component, and somehow it's rendered multiple times (e.g. Suspense), will there be a problem?

src/style.js
giuseppeg: Sounds good!

src/style.js
giuseppeg: The first time it is called `add` injects the styles into an in memory stylesheet. Future calls with the same props will just increment a counter

https://github.com/vercel/styled-jsx/blob/27418790cc3031113d031e2742ecdaab197e18a3/src/stylesheet-registry.js#L55-L58

styled-jsx won't work with streaming rendering at the moment and perhaps suspense. I think it'd need a way to yield styles for the page while preserving a global state (registry) to avoid duplicates. I am open to collaborations if you'll ever need help with this.

huozhi: Tested with suspense in react 18, looks good so far. As the [anncoucement](https://github.com/reactwg/react-18/discussions/7) said effects hooks are reliable with suspense.

shuding: > If we SSR the component, and somehow it's rendered multiple times (e.g. Suspense), will there be a problem?

To add more context to this, normally when we SSR all components will only be rendered once and there's no re-render. But with Suspense in the server side, it could be possible for a component to be re-rendered multiple times due to Suspense fallback and retry. In that case the counter will be wrongly counted. However since we won't have unmount in SSR and there's no style removal, so I think it's OK but I can't be 100% sure.

With CSR Suspense it won't be an issue though since we have every side effect inside the `useLayoutEffect` hook.

We will figure this out. 👍

giuseppeg: > I use the streaming API with next.js to test it locally 

@huozhi interesting, I'd be curious to see how this works but in general I was thinking that as you stream HTML you'd need to flush styles (render style tags) but keep an in memory registry on the server so that you don't accidentally render the same styles in future chunks (imagine you have 4 instances of a Button component).

Probably this could be hacked with the current utils but styled-jsx doesn't do this out of the box at the moment.

huozhi: I can give an example that extra style rules are injected on server side. i.e. As you said we have 4 buttons, and we suspend those buttons to let them render on client. Server side only emits the fallbacks.

For suspensed components they'll only send some placeholders, like

```html
<!--$?--><template id="B:0"></template>Loadding...<!-- -->
<!--$?--><template id="B:1"></template>Loadding...<!-- -->
...
```

It also emits some scripts inside the first screen html, to make it simple I'll define the function name in my way.

```html
<script>
function refreshSuspense(id) { /* load suspenses on client side */  }
</script>
<!-- then refresh the placeholders on client to get the final content -->
<script>refreshSuspense('B:0')</script>
<script>refreshSuspense('B:1')</script>
```

The bad thing here is on server side, `JSXStyle` doesn't know if it will be suspended or not. It will inject its style to registry anyway. So the first screen content are fallbacks with **1 style rule injected globally**. But that rule won't be useful until the button's content is fetched.

giuseppeg: `JSXStyle` should be a direct descendant of `Button` so if you render a placeholder for `Button`, styles won't be registered until later when on the client you call `refreshSuspense` and render `Button`. 

The only case where you could send styles for suspended elements is when React renders something but then throws it away (does not "commit" it and sends a placeholder instead) but I am not even sure if they do this kind of thing during SSR. 

Anyway all the above are assumptions, I am not sure how you or the React team is implementing SSR with Suspense but you definitely have more insights to make the call so consider this PR approved if any of your teammates gives you a thumb up.

huozhi: Just checked with team that's the prototype behavior which wait all html content resolved on server side. So that styles get injected. The expected first stream sent to client shouldn't contain styles in that case. Will do more tests to see how to handle it gracefully.

devknoll: Note: streaming means multiple renders can happen concurrently, so a global registry on the server is unworkable. We need to be able to provide a registry per request (as in #703) and get a list of styles that we can deduplicate or is deduplicated already

shuding: > Note: streaming means multiple renders can happen concurrently, so a global registry on the server is unworkable with Concurrent Mode.

I wonder why is that unworkable. When doing SSR, the registry doesn't need to track the order of renders or concurrency, it just need to work like a boolean (only add the style when the counter turns from 0 to 1) since there's no removal process.

Or does it mean multiple requests, when you say "multiple renders" here?

devknoll: Yes, multiple requests can be being processed concurrently, so a singleton that tracks styles used wouldn’t work on the server, as other requests would read/write the same singleton.

devknoll: This works in synchronous React because the render starts and ends synchronously. But in React 18 a tree could suspend while waiting for data.

giuseppeg: Is it possible for the following race condition to happen?

1. big subtree starts rendering a Button registers the styles
2. smaller subtree renders Button and is ready before 1.
3. the Button from 2. is unstyled

By the way a while ago I created a gist / PoC for a Style component that should work with streaming/suspense - I'll see if I can polish it and share it in the following days. As @devknoll mentioned it uses Context

shuding: Yeah... if that "smaller subtree renders Button and is ready before 1." is inside another Suspense boundary, it’s gonna be unstyled until it’s hydrated.

src/style.js
huozhi: observed some interesting transpile behavior only on **production** here:
`children` has been transformed to strings concatenation  🤔 then `JSXStyle` always get children as array so it will keep triggering hooks

```js
o(
  r.default, {
    id: '200133253',
    dynamic: [s, i],
  },
  [
    '.sth.__jsx-style-dynamic-selector{color:'
    .concat(s, ';background:')
    .concat(
      i,
      ';display:inline-block;font-weight:',
    )
    .concat(u ? 'bold' : 'normal', ';padding:')
  ]
)
```

giuseppeg: Yes it looks like a template literal with expressions, weird that it is evaluated and inlined. Anyway it should be ok to remove children from the deps!

package.json
```diff
@@ -91,7 +91,7 @@
     "xo": "0.24.0"
   },
   "peerDependencies": {
-    "react": "15.x.x || 16.x.x || 17.x.x"
+    "react": "17.x.x || 18.x.x"
```
giuseppeg, line 94: why are we dropping 16.x.x?

giuseppeg (approved):
Check the peerDependencies comment and then feel free to merge. Remember to add the comment below when squashing

BREAKING CHANGE: This change drops support for React 15.

package.json
huozhi: it's because of usage of hooks. Tested the implementation based on react 16 class components lifecycles that doesn't work well with this effect. After switching to `useLayoutEffect` the style registry handles it properly. So this would be a breaking change that dropping both react 15 and 16 😿

giuseppeg: So `useLayoutEffect` behaves differently in 17+?

huozhi: Oh it should include `>= 16.8`, updated.

giuseppeg:
@huozhi I am thinking that since there will be more breaking changes in the other PRs maybe we can create a canary release and once we merge all the changes necessary to support React18 we can cut a major release

huozhi:
@giuseppeg yeah totally, @shuding and I were thinking the same thing. I'll update the outdated `cancary` branch to support canary release. 🙏

giuseppeg (approved):

PR #721, opened 2021-07-22
feat: contextual styled-jsx

huozhi:
### Feature

This is the draft for contextual styled-jsx rendering idea from https://gist.github.com/giuseppeg/dc4cecdae1e9fcb4c5761712e75e9b93

Which would be a benefit for multi-requests case with concurrent mode on server side, metioned in #720 .
Each request will only use its own context to add/remove styles, to avoid having a shared style registry instance messed up on server side.

Use case

```jsx
import React from 'react'
import ReactDOM from 'react-dom/server'
import { StyleRegistry, useStyleRegistry } from 'styled-jsx'
import App from './app'

function Head() {
  const registry = useStyleRegistry()
  const styles = registry.styles()
  // in blocking rendering, flush styles in each render
  registry.flush()
  return <head>{styles}</head>
}

export default (req, res) => {
  const app = ReactDOM.renderToString(<App />)
  const html = ReactDOM.renderToStaticMarkup(
    <StyleRegistry>
      <html>
        <Head />
        <body>
          <div id="root" dangerouslySetInnerHTML={{ __html: app }} />
        </body>
      </html>
    </StyleRegistry>
  )
  res.end('<!doctype html>' + html)
}
```

Closes #703 
Closes #485 
Closes #563


huozhi:
Draft the idea for moving style registry to react context for safety in concurrent mode. The APIs are not stabilized, feel free to share your opinions about it.
@giuseppeg @shuding @devknoll

test/babel6/stylesheet-registry.js
```diff
@@ -2,7 +2,7 @@
 import test from 'ava'
 
 // Ours
-import StyleSheetRegistry from '../../src/stylesheet-registry'
+import { StyleSheetRegistry } from '../../src/stylesheet-registry'
```
giuseppeg, line 5: Since this is going to be a major release probably we can drop support for Babel 6

test/index.js
```diff
@@ -187,6 +187,7 @@ test('server rendering', t => {
   t.is('', flushToHTML())
 })
 
+/*
```
giuseppeg, line 190: Why are you commenting these tests?

src/server.js
```diff
@@ -1,6 +1,11 @@
 import React from 'react'
-import { flush } from './style'
 
+// fake flush
+function flush() {
+  return []
+}
+
+/* @deprecated */
```
giuseppeg, line 8: I would fail here or warn otherwise it will "fail" silently i.e. `flushToReact` and `flushToHTML` continue to work but output nothing

src/style.js
```diff
@@ -1,57 +1,26 @@
...
+  if (typeof window === 'undefined') {
+    registry.add(props)
     return null
   }
+  useLayoutEffect(() => {
+    registry.add(props)
+    return () => {
+      registry.remove(props)
+    }
+  }, [props.id, props.children, String(props.dynamic)])
```
giuseppeg, line 14: Registry is a dependency too - though if the registry is swapped everything will break.

giuseppeg (commented):
Thanks for putting this together @huozhi 

I am afraid that the moment we introduce a configurable registry via context we open up the possibility to get into this issue https://github.com/vercel/styled-jsx/pull/703#issuecomment-811863907 i.e. if you have multiple registries in the page the first one will  pick up all the server side rendered style tags regardless of where the came from (which registry they belong to).

In general if you want styled-jsx to work with modern React you'll have to change quite a few things and it would be worth considering all the constraints and edge cases together (concurrent mode, streaming rendering, multiple registries/context etc). 

Unfortunately I don't have time to do all this work on my spare time but if you all put together a detailed proposal/rfc I'd be happy to take a look.

fwiw the gist you found has flaws but I haven't had the time to iterate on it - I just made it public because I thought nobody would find it :)

src/style.js
```diff
@@ -1,35 +1,29 @@
...
-      styleSheetRegistry.remove(props)
+      registry.remove(props)
     }
     // props.children can be string[], will be striped since id is identical
   }, [props.id, String(props.dynamic)])
   return null
 }
 
 JSXStyle.dynamic = info => {
+  const registry = useContext(StyleSheetContext)
```
giuseppeg, line 21: I am afraid that when you have an early return in a component this will result in an error. See https://codesandbox.io/s/ecstatic-darkness-8yd0e?file=/src/App.js

Screen Shot 2021-08-11 at 9 48 05 AM

giuseppeg: oh nevermind probably that's never going to be the case because the Babel plugin adds the `JSXStyle.dynamic` to every JSX element

giuseppeg: `registry.computeId` and `registry.computeSelector` are pure functions so you could move them outside of the `StyleSheetRegistry` class and have them share a global cache. This would allow you to remove the context dependency from `JSXStyle.dynamic`

huozhi: Nice callcout! I think the cache of `registry.computeId` and `registry.computeSelector` should live per request like the style registry context, to avoid conflicts in concurrent mode? Might be bit hard to do if they don't consume cache or registry from context?

giuseppeg: The cache can be shared because it is there just to speed things up a bit. There won't be conflicts

https://github.com/vercel/styled-jsx/blob/d591ce27c9ae4d41d7833551e5a153324cdd69ca/src/stylesheet-registry.js#L134-L150

You can turn this into an IIFE, or just define the cache as a local variable in the module

devknoll: I wonder if we can rename `JSXStyle.dynamic` to `JSXStyle.useDynamic` since it's technically a hook now

giuseppeg: I am suggesting to remove the dependency from context because it is not needed to compute the ids. We just need to move `computeId` and `computeSelector` outside of the `StyleSheetRegistry` class and export them.

giuseppeg: I mean we don't need those factories `createComputeId` and `createComputeSelector` and one version of the resulting function per `StyleSheetRegistry` instance. The two caches can be shared

src/stylesheet-registry.js
```diff
@@ -163,11 +183,17 @@ function invariant(condition, message) {
...
+export function createStyleRegistry() {
+  return new StyleSheetRegistry()
+}
 
-export function StyleRegistry({ children }) {
+export function StyleRegistry({ registry: configuredRegistry, children }) {
   const rootRegistry = useContext(StyleSheetContext)
-  const [registry] = useState(() => rootRegistry || new StyleSheetRegistry())
+  const [registry] = useState(
+    () => configuredRegistry || rootRegistry || createStyleRegistry()
```
shuding, line 195: Since we only want 1 registry to be created globally, shouldn’t rootRegistry always take priority here?

src/stylesheet-registry.js
```diff
@@ -179,31 +205,5 @@ export function StyleRegistry({ children }) {
...
-      id: `__${id}`,
-      // Avoid warnings upon render with a key
-      key: `__${id}`,
-      nonce: options.nonce ? options.nonce : undefined,
-      dangerouslySetInnerHTML: {
-        __html: css
-      }
-    })
-  })
+  return useMemo(() => registry, [registry])
```
shuding, line 208: This seems unnecessary.

src/stylesheet-registry.js
```diff
@@ -163,11 +183,17 @@ function invariant(condition, message) {
   }
 }
 
-export const StyleSheetContext = createContext(new StyleSheetRegistry())
+export const StyleSheetContext = createContext(null)
+
+export function createStyleRegistry() {
```
shuding, line 188: Why an extra function as we already exported the class

huozhi: `StyleSheetRegistry` export is for testing. Every valid imports will come from index later.

readme.md
```diff
@@ -689,21 +687,26 @@ Styles can be preprocessed via plugins.
 Plugins are regular JavaScript modules that export a simple function with the following signature:
 
 ```js
```
shuding, line 689:
```suggestion
```ts
```

readme.md
```diff
@@ -689,21 +687,26 @@ Styles can be preprocessed via plugins.
...
-(css: string, options: Object) => string
+;(css: string, options: Object) => string
```
shuding, line 690:
```suggestion
(css: string, options: Object) => string
```

readme.md
```diff
@@ -339,48 +334,32 @@ Please keep in mind that constants defined outside of the component scope are tr
...
-  const app = ReactDOM.renderToString(<App />)
-  const styles = flush()
-  const html = ReactDOM.renderToStaticMarkup(<html>
-    <head>{ styles }</head>
-    <body>
-      <div id="root" dangerouslySetInnerHTML={{__html: app}} />
-    </body>
-  </html>)
-  res.end('<!doctype html>' + html)
+function Head() {
```
shuding, line 343:
```suggestion
function HeadStyles() {
```

To prevent confusion between this and `next/head`.

readme.md
```diff
@@ -339,48 +334,32 @@ Please keep in mind that constants defined outside of the component scope are tr
...
-      <head>${styles}</head>
-      <body>
-        <div id="root">${app}</div>
-      </body>
-    </html>`
-  res.end(html)
+  const html = ReactDOM.renderToStaticMarkup(
+    <StyleRegistry>
+      <html>
+        <Head />
```
shuding, line 355:
```suggestion
        <HeadStyles/>
```

shuding (approved):

huozhi:
going to merge this PR into alpha branch to create a prerelease now. any further concern with this PR please leave comments I'll address them later 🙏 cc @giuseppeg

PR #725, opened 2021-08-03
fix: css typing

huozhi:


css.d.ts
```diff
@@ -1,11 +1,11 @@
...
-declare function css(chunks: TemplateStringsArray, ...args: any[]): JSX.Element
+declare function css(chunks: TemplateStringsArray, ...args: any[]): string
 declare namespace css {
-  function global(chunks: TemplateStringsArray, ...args: any[]): JSX.Element
+  function global(chunks: TemplateStringsArray, ...args: any[]): string
   function resolve(
     chunks: TemplateStringsArray,
     ...args: any[]
-  ): { className: string; styles: JSX.Element }
+  ): { className: string; styles: string }
```
giuseppeg, line 9: In this case `styles: JSX.Element` is correct

giuseppeg (approved):

PR #726, opened 2021-08-03
refactor: replace import helper

huozhi:
Use import expression injection to replace with `@babel/helper-module-imports`

Currently styled-jsx has inconsistence behavior between `styled-jsx/style` and `styled-jsx/css` imports.

`addDefault` helper will emit `styled-jsx/style` code based on babel `sourceType`, but `resolve` tags imported from `styled-jsx/css` aren't handled if it's in `sourceType="script"` mode

For example, the code below won't be transpiled

```js
const css = require('styled-jsx/css')

exports.normal = css.resolved`p { color: blue }`
exports.resolved = css`p { color: red }`
```

But any `<style jsx>` usage in commonjs module will be transpiled and inject the imports/require.

Import declarations can be easily captured in AST traversal, but require expressions (w/ or w/o module interop) is hard to be captured. Such as`styled-jsx/css` is using this way to transpile `css` related usages.

It's better to only support ESM transform and let other plugin for example (commonjs plugin) to handle the ESM to CJS transform.

### Suggested Usage

Use ESModule syntax for jsx styles. If you want to transformed to CJS, use commonjs plugin


src/babel.js
```diff
@@ -288,16 +289,20 @@ export default function({ types: t }) {
...
-          )
+          // create unique identifier for _JSXStyle component
+          state.styleComponentImportName = path.scope.generateUidIdentifier(
+            STYLE_COMPONENT
+          ).name
+          const importDeclaration = createReactComponentImportDeclaration(state)
+          state.styleImportPath = path.unshiftContainer(
+            'body',
+            importDeclaration
+          )[0]
```
giuseppeg, line 305: Now we can move this to `Program.exit` and avoid to insert it here to then remove it at the end if it was not needed

huozhi: I did this because if it's not been used it could be removed at exit after other transform.

e.g.

```
inject imports
transform to cjs by other plugin
detect usage (not used)
remove tranfromed imports (cjs require)
```

If we directly inject at exit, commonjs plugin doesn't take effect

huozhi: Updated. I thought it won't work with cjs plugin, turns out it could.

giuseppeg (approved):

PR #730, opened 2021-08-06
perf: drop babel 6 support

TrySound:
It's been a while since babel 7 is out. I think it makes sense to get
rid from old babel dependencies and bump to the latest version.


giuseppeg:
Thanks @TrySound ! we are preparing a release with breaking changes and temporarily merging to `alpha` can you change the base branch?

huozhi (approved):
@TrySound Thank you!

Issue #733, opened 2021-08-16
Support ESM import syntax for plugins


Divlo:
#### What is the current behavior?

Currently, `styled-jsx` plugins can't be written with ESM imports syntax because it requires the import instead of using [dynamic import](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_Imports) syntax.

The problematic line of code is: https://github.com/vercel/styled-jsx/blob/e41c6fe488b7c25ddc2a8033106f1df424986ccc/src/_utils.js#L530

#### What is the expected behavior?

`styled-jsx` should be able to work with plugins written with ESM.

Related: https://github.com/Thream/styled-jsx-plugin-sass/issues/68

PR #735, opened 2021-08-17
chore: use modern eslint and limit engines

huozhi:
* replace `xo` with modern `eslint`
* replace `mz` with `fs/promises`
* require engine node >= 10.0

This can be precondition for unlock ESM module support requirement in #733

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "react": ">= 16.8.0 || 17.x.x || 18.x.x"
   },
   "engines": {
-    "node": ">= 10"
+    "node": ">= 12.0.0"
   },
   "keywords": [
     "babel-plugin-macros",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -78,19 +81,21 @@
...
     "react-dom": "17.0.1",
-    "semantic-release": "17.2.2",
-    "xo": "0.24.0"
+    "semantic-release": "17.2.2"
   },
   "peerDependencies": {
     "react": ">= 16.8.0 || 17.x.x || 18.x.x"
   },
+  "engines": {
+    "node": ">= 10"
```
giuseppeg, line 97: I think that we can drop support for 10 (and probably 12). What does Next.js support?

giuseppeg (dismissed):

package.json
huozhi: switched to `node >= 12.0.0`, aligning with next.js

giuseppeg (approved):

PR #739, opened 2021-08-27
fix: mark @babel/core as optional peer dependency

Unbuttun:
Fixes The missing dependency error that comes while installing it through `Yarn 2, Yarn 3`.
 and also fixes https://github.com/vercel/styled-jsx/issues/736

Error - 
`
styled-jsx@npm:4.0.0 [78a22] doesn't provide @babel/core (pad17e), requested by @babel/plugin-syntax-jsx
`

The user needs to add 
```yml
packageExtensions:
  styled-jsx@*:
    dependencies:
      "@babel/core": "*"
 ```
in .yarnrc.yml  even when they're not using this dependency directly, for example while using next.js

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -59,6 +59,7 @@
     }
   },
   "dependencies": {
+    "@babel/core": "7.12.3",
```
huozhi, line 62: This jsx plugin should use `@babel/core` on user side so we don't want it to become dependency.

We can add the meta field to package.json to declare it as optional

```json
"peerDependenciesMeta": {
  "@babel/core": {
    "optional": true
  }
}
```

Unbuttun: Thanks for the, did not knew this option existed.

huozhi (approved):

huozhi:
Thanks!

0899279111:
h

0899279111 (approved):

PR #747, opened 2021-09-06
[WIP][COMPARISON] upgrade stylis

huozhi:
This is a draft PR to check the effects of upgrading to latest stylist

- [ ] Understand reason of duplicated class names generated
- [ ] Disabling nesting


ywroh:
The reason that the duplicate class name is created is that the element was added redundantly in the namespace, which is stylis middleware.

I modified this content and made a pull request to the stylis repository as follows.

[https://github.com/thysultan/stylis.js/pull/282](url)

huozhi:
@ywroh looks like it's by designed (https://github.com/thysultan/stylis.js/issues/101)

PR #749, opened 2021-09-15
fix: Make `JSXStyle` return a noop if the registry context is not provided

shuding:
It is possible to (fully or partially) render a component without the context. One legit use case is to generate and parse the markup without rendering it in the browser (so no registry is needed). In those cases we can simply ignore all `JSXStyle` elements.

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -1,12 +1,19 @@
-import { useLayoutEffect, useContext } from 'react'
-import { StyleSheetContext } from './stylesheet-registry'
+import { useLayoutEffect } from 'react'
+import { useStyleRegistry } from './stylesheet-registry'
 import { computeId } from './lib/hash'
 export default function JSXStyle(props) {
-  const registry = useContext(StyleSheetContext)
+  const registry = useStyleRegistry()
+
+  // If `registry` does not exist, we do nothing here.
+  if (!registry) {
+    return null
+  }
+
   if (typeof window === 'undefined') {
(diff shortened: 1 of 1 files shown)


timneutkens (approved):
Thanks! 🚀

src/style.js
```diff
@@ -1,19 +1,27 @@
...
+import { useLayoutEffect } from 'react'
+import { useStyleRegistry } from './stylesheet-registry'
 import { computeId } from './lib/hash'
 export default function JSXStyle(props) {
-  const registry = useContext(StyleSheetContext)
+  const registry = useStyleRegistry()
+
+  // If `registry` does not exist, we do nothing here.
+  if (!registry) {
+    return null
```
giuseppeg, line 9: The following is just an observation and me asking out of curiosity, not a big deal:

Normally I throw an error here to help users understand what's going on. In the case you described in the PR description can't you remove the babel plugin altogether so that style tags are not turned into JSXStyle elements?

giuseppeg: this is what we have for test environments https://github.com/vercel/styled-jsx/blob/master/src/babel-test.js

shuding: In certain use cases, for example a lib that renders a component (which has styled-jsx inside) with `renderToStaticMarkup`. We can't throw since it should not block rendering, and we can't ask them to disable the babel plugin as well because it's not a project level thing.

PR #750, opened 2021-09-23
docs: Add TypeScript config instructions to readme

Macil:
This recommended configuration is inspired by Next.js's own typescript template which contains a [file](https://github.com/vercel/next.js/blob/canary/packages/create-next-app/templates/typescript/next-env.d.ts) with a similar triple-slash directive that transitively imports styled-jsx's types.


huozhi (approved):
Thanks!

PR #766, opened 2021-12-17
fix: use string for nonce typing

huozhi:
React warning:
```
If you want to write it to the DOM, pass a string instead: nonce="true" or nonce={value.toString()}
```

we need to use the string value for dom property `nonce`

x-ref: https://github.com/vercel/next.js/pull/32594


index.d.ts
```diff
@@ -8,7 +8,7 @@ declare module 'react' {
 }
 
 export type StyleRegistry = {
-  styles(options?: { nonce?: boolean }): JSX.Element[]
+  styles(options?: { nonce?: string | boolean }): JSX.Element[]
```
ijjk, line 11: What is the expected behavior when `nonce: true` is passed?

huozhi: `nonce="true"` boolean is still coverted to string, we can also change to string if we don't need the back compatible since it's bit trivial

test/index.js
```diff
@@ -20,7 +20,7 @@ const flushToHTML = (registry, options = {}) => {
     const id = args[0]
     const css = args[1]
     html += `<style id="__${id}"${
-      options.nonce ? ` nonce="${options.nonce}"` : ''
+      options.nonce ? ` nonce="${options.nonce.toString()}"` : ''
```
ijjk, line 23: Does this need to be added below in `mapCssRulesToReact` as well?

index.d.ts
ijjk: Should `nonce="true"` be rendered to the DOM, doesn't the nonce need to match the header/meta tag to be valid?

PR #768, opened 2021-12-20
perf: fallback to module level registry in browser

huozhi:
Fallback to module level local registry on client side, so we don't have to include a registry there.
`StyleRegistry` is only required for server render for safe concurrent rendering

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -1,19 +1,22 @@
 import React from 'react'
-import { useStyleRegistry } from './stylesheet-registry'
+import { useStyleRegistry, createStyleRegistry } from './stylesheet-registry'
 import { computeId } from './lib/hash'
 
 // Opt-into the new `useInsertionEffect` API in React 18, fallback to `useLayoutEffect`.
 // https://github.com/reactwg/react-18/discussions/110
 const useInsertionEffect = React.useInsertionEffect || React.useLayoutEffect
+
+const isBrowser = typeof window !== 'undefined'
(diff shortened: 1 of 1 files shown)


src/style.js
```diff
@@ -1,19 +1,22 @@
...
-import { useStyleRegistry } from './stylesheet-registry'
+import { useStyleRegistry, createStyleRegistry } from './stylesheet-registry'
 import { computeId } from './lib/hash'
 
 // Opt-into the new `useInsertionEffect` API in React 18, fallback to `useLayoutEffect`.
 // https://github.com/reactwg/react-18/discussions/110
 const useInsertionEffect = React.useInsertionEffect || React.useLayoutEffect
+
+const isBrowser = typeof window !== 'undefined'
+const defaultRegistry = createStyleRegistry()
```
timneutkens, line 10:
```suggestion
const defaultRegistry = typeof window !== 'undefined' ? createStyleRegistry() : undefined
```

src/style.js
```diff
@@ -1,19 +1,22 @@
...
 export default function JSXStyle(props) {
-  const registry = useStyleRegistry()
+  const registry = isBrowser ? defaultRegistry : useStyleRegistry()
```
timneutkens, line 12:
```suggestion
  const registry = defaultRegistry ? defaultRegistry : useStyleRegistry()
```

timneutkens (approved):

PR #769, opened 2022-01-11
fix: isBrowser false positives

AaronO:
Feature detect against `document` instead of `window` to avoid false positives in environments like Deno that define a `window` global but no `document`


huozhi:
Hi @AaronO could you provide an example that how you use styled-jsx with deno?

PR #770, opened 2022-01-14
perf: Pre-compile dependencies to reduce install size/time

ijjk:
This moves all dependencies to `devDependencies` by pre-compiling with `ncc` to reduce the install size and time. 

Before: `2.5M	styled-jsx-pre-ncc/node_modules`
After: `904K	styled-jsx-post-ncc/node_modules`

This removes 17 nested dependencies.

x-ref: https://github.com/vercel/next.js/pull/32742

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules
 
 # build output
 dist
+out
 
 # logs
 npm-debug.log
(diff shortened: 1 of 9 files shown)


huozhi (approved):
🚀

huozhi:
We might can also get rid of the warning of styled-jsx when install next.js, since babel plugin is not a dep anymore

PR #802, opened 2022-07-17
build: use bunchee for bundling

huozhi:
[bunchee](https://www.npmjs.com/package/bunchee) is a zero configuration bundler based on rollup and swc, it can get rid of the webpack runtime code which relying on nodejs runtime variable like `__dirname` ncc brought in

Also faster ⚡️

```
✨ Built styled-jsx
✅ Finished in 900 ms
$ bunchee src/webpack.js -f cjs --runtime node -o dist/webpack/index.js
✨ Built styled-jsx
✅ Finished in 226 ms
$ bunchee src/index.js -f cjs --runtime node -e react -o dist/index/index.js
✨ Built styled-jsx
✅ Finished in 63 ms
✨  Done in 2.52s.
```

x-ref: https://github.com/vercel/styled-jsx/issues/784
x-ref: https://github.com/vercel/next.js/pull/40165


src/babel.js
```diff
@@ -18,13 +16,8 @@ import {
...
 
-export function macro() {
-  return require('./macro')
-}
-
-export function test() {
-  return require('./babel-test')
-}
+export { default as macro } from './macro'
+export { default as test } from './babel-test'
```
ijjk, line 20: Should these still be functions that export these? E.g.

```js
import _macro from './macro'

export function macro() {
  return _macro
}
```

ijjk (approved):

PR #805, opened 2022-08-08
fix: declare module and fix conflict typing name

huozhi:
x-ref: https://github.com/vercel/next.js/runs/7729014049?check_suite_focus=true

Add declare module and change type name from `StyleRegistry` to `StyledJsxStyleRegistry`

follow-up: will change the styled-jsx assets copy script in next.js


ijjk (commented):
Should we be using `declare module` here? If it's needed here I think we would need it for the other type declarations as well

PR #806, opened 2022-08-10
fix: tweak types

huozhi:
x-ref: https://github.com/vercel/next.js/pull/39474

`declare module` wasn't picked up by typescript

* Add `global.d.ts`
* Move react style types overriding into `global.d.ts`
* Scope react types in `index.d.ts`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,13 +1,6 @@
-import React from 'react'
-
-declare module 'react' {
-  interface StyleHTMLAttributes<T> extends HTMLAttributes<T> {
-    jsx?: boolean
-    global?: boolean
-  }
-}
-
 declare module 'styled-jsx' {
+  import React from 'react'
+
   export type StyledJsxStyleRegistry = {
     styles(options?: { nonce?: string }): JSX.Element[]
     flush(): void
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -1,13 +1,6 @@
...
-
-declare module 'react' {
-  interface StyleHTMLAttributes<T> extends HTMLAttributes<T> {
-    jsx?: boolean
-    global?: boolean
-  }
-}
-
 declare module 'styled-jsx' {
+  import React from 'react'
```
ijjk, line 2: Should we use `import('react').ReactNode` below instead?

huozhi: They both work, but `import React` will only work when I moved it into the styled-jsx scope. Seems it does have side effect?
I was just curious why you use `import('react')` cuz if there're more used types from react we need to write more `import()` so I tried this. I can change to `import('react')` since there's only 1 used type now.

ijjk: Yeah does seem there's a side-effect there, was mainly thinking to use it to limit potential conflicts

ijjk (approved):

PR #812, opened 2022-09-05
fix: load babel macro dynamically

huozhi:
Use `require` call to dynamic require macro plugins, so that it won't be bundled in the dist/babel assets when you're not using macro

* chore: Upgrade bunchee to latest
* test: create babel macro test module for testing


src/macro.js
```diff
@@ -1,11 +1,12 @@
-import { createMacro, MacroError } from 'babel-plugin-macros'
 import { processTaggedTemplateExpression } from './babel-external'
 import {
   setStateOptions,
   createReactComponentImportDeclaration
 } from './_utils'
 import { STYLE_COMPONENT } from './_constants'
 
+const { createMacro, MacroError } = (() => require('babel-plugin-macros'))()
```
ijjk, line 8: Won't this still be called whenever `styled-jsx/babel` is loaded since it's imported unconditionally here? https://github.com/vercel/styled-jsx/blob/2a31e7395f76bb997909734cf59a0abc7f4bbdd9/src/babel.js#L19

ijjk (approved):

PR #816, opened 2022-10-09
feat: styled-jsx as client-only for rsc

huozhi:
Use `import "client-only"` to mark styled-jsx as client only packages for server components, so that when users try to import it on server layer it will error in next.js

Expected error if you use styled-jsx in server components rather than client components
> error - ../../../node_modules/.pnpm/client-only@0.0.1/node_modules/client-only/error.js (1:6) @ eval
error - Error: This module cannot be imported from a Server Component module. It should only be used from a Client Component.
    at eval (webpack-internal:///(sc_server)/../../../../node_modules/.pnpm/client-only@0.0.1/node_modules/client-only/error.js:1:7)

x-ref: https://github.com/vercel/next.js/pull/41414

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "ava": "4.3.1",
     "babel-plugin-macros": "2.8.0",
     "bunchee": "2.1.5",
-    "client-only": "0.0.1",
     "convert-source-map": "1.7.0",
     "eslint": "7.32.0",
     "eslint-config-prettier": "4.0.0",
@@ -123,5 +122,8 @@
     "zeit",
     "css-in-js",
     "css"
-  ]
+  ],
+  "dependencies": {
+    "client-only": "0.0.1"
+  }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -122,5 +122,8 @@
     "zeit",
     "css-in-js",
     "css"
-  ]
+  ],
+  "dependencies": {
+    "client-only": "0.0.1"
```
ijjk, line 127: Does this need to be a runtime dependency instead of bundling?

huozhi: Yeah it has to be so that if they occurred in the server bundle graph we'll error (https://www.notion.so/vercel/React-Server-Condition-Poisoning-aa31aa23e26d41a18b4450725d37a563)

ijjk (approved):

vpontis:
Ooo exciting! Congrats on 5.1.0 👍👍

PR #828, opened 2023-03-07
Add test case for `@container` query

vpontis:
This test fails. I think it's because `@container` is only supported in the most recent version of Stylis ([issue](https://github.com/thysultan/stylis/issues/303)) but styled-jsx uses an old version of Stylis. 

I'm looking into how hard it would be to update Stylis to the newest version...


testqer09:
test

testqer09:
https://qertest1234.atlassian.net/browse/TREE-1

btaganus:
https://qertest1234.atlassian.net/browse/TREE-1

Issue #556, opened 2019-02-24
"Cannot read property 'name' of undefined" error from isStyledJsx in _utils.js


TxHawks:
#### Do you want to request a *feature* or report a *bug*?
Report a bug

#### What is the current behavior?
When using Next 8.0.1 and styled-js 3.2.1 (installed as a next dependency), I get a `Cannot read property 'name' of undefined` TypeError in `_utils.js`.  This only happens when also using [`next-transpile-modules`](https://github.com/martpie/next-transpile-modules)

The offending code is the check for `attr.name.name` in line 124:
https://github.com/zeit/styled-jsx/blob/00fcb7bcad9c8af91167056f93c63858d24769ba/src/_utils.js#L121-L124

We don't actually use styled-jsx, and I have no idea our app is running into this, but patching it like this avoids the issue:
```js
el.openingElement.attributes.some(attr => attr.name && attr.name.name === STYLE_ATTRIBUTE)
```

If deemed worthy, I'm happy to send a pull request with the above change.

#### Environment (include versions)

* OS: osx 10.13.6
* Browser: compoile time error
* styled-jsx (version): 3.2.1

#### Did this work in previous versions?
I don't know

giuseppeg:
hi this could be a precedence issue i.e. maybe `next-transpile-modules` runs before styled-jsx and it is transpiling something which leads to the bug. Have you tried to remove styled-jsx from the next/babel preset? What does your `transpileModules` look like?

TxHawks:
Sorry for the long delay in replying. Closing since the issue seem to no longer happen with latest version of Next.

PR #843, opened 2024-02-29
Bugfix: handle undefined <style> tag attr.name

LucasMatuszewski:
I've faced an issue when `attr.name` was undefined.
For example when spreading an object inside the `<style>` tag like this:
```
<style {...(key && { key })} />
```
I got an error on the Storybook build:
```
Module build failed (from ./node_modules/babel-loader/lib/index.js):
TypeError: /home/lucas/project-name/src/utils/withStyles.tsx: Cannot read properties of undefined (reading 'name')
    at /home/lucas/project-name/storybook/node_modules/styled-jsx/dist/babel/index.js:21535:26
```
And in the debugger:
```
Exception has occurred: TypeError: Cannot read properties of undefined (reading 'name')
  at /home/lucas/DEV/project-name/storybook/node_modules/styled-jsx/dist/babel/index.js:21535:26
```

I have applied the same fix as the one in `isGlobalEl` function from `JJ Kasper (January 6th, 2019 6:09 PM)`.

Sorry, if I should create an issue first ;)

### DIFF
--- a/src/_utils.js
+++ b/src/_utils.js
@@ -124,7 +124,7 @@ export const isGlobalEl = el =>
 export const isStyledJsx = ({ node: el }) =>
   t.isJSXElement(el) &&
   el.openingElement.name.name === 'style' &&
-  el.openingElement.attributes.some(attr => attr.name.name === STYLE_ATTRIBUTE)
+  el.openingElement.attributes.some(({ name }) => name && name.name === STYLE_ATTRIBUTE)
 
 export const findStyles = path => {
   if (isStyledJsx(path)) {


seanleecoder:
Independent reproduction from a different toolchain than the original report — this is still live in 5.1.7 and on `main`.

**Trigger:** a Next.js app (16.1.6, Babel rather than SWC) that lists `expo` in `transpilePackages`. That makes Next's Babel pipeline — styled-jsx plugin included — run over the dependency's JSX source, which contains:

```jsx
// expo/src/launch/AppRegistry.web.tsx
<style {...props} />
<style {...props} dangerouslySetInnerHTML={{ __html: sheet.textContent }} id={sheet.id} />
```

`isStyledJsx` iterates that element's attributes and hits a `JSXSpreadAttribute`, which carries `argument` rather than `name` — so `attr.name.name` throws.

Two things that make this harder to attribute than a normal crash, and are worth recording for whoever triages it:

1. **The application contains no `<style jsx>` at all.** Every `<style>` element in our own source uses explicit attributes; the crash comes entirely from a transpiled dependency. So the stack trace points at styled-jsx in a codebase that never knowingly uses styled-jsx.
2. **Any `transpilePackages` (or the older `next-transpile-modules`) setup is exposed** as soon as one transpiled dependency spreads props onto a `<style>` element — and that dependency can add such a line at any time.

That combination is probably why [#556](https://github.com/vercel/styled-jsx/issues/556) (2019, also via `next-transpile-modules`) proposed this exact one-line patch, then got closed by the reporter with "seems to no longer happen" rather than by the fix landing. The reporter there also said "we don't actually use styled-jsx, and I have no idea why our app is running into this."

**The existing asymmetry is an argument for merging as-is:** `isGlobalEl`, two lines above, already guards in precisely this shape —

```js
export const isGlobalEl = el =>
  el && el.attributes.some(({ name }) => name && name.name === GLOBAL_ATTRIBUTE)
```

so this PR makes `isStyledJsx` consistent with its immediate neighbour rather than introducing a new pattern.

We're carrying this as a package patch in the meantime. Happy to verify against a release candidate.

Issue #856, opened 2025-04-28
Cannot read properties of undefined (reading 'appendChild')


JeremyRH:
Steps to reproduce:
- Use `<style jsx>`.
- Use React 18+.
- [Pass `document` as the first argument to ReactDOM client's `hydrateRoot`](https://react.dev/reference/react-dom/client/hydrateRoot#hydrating-an-entire-document).
- [Have a hydration mismatch between server and client](https://react.dev/reference/react-dom/client/hydrateRoot#:~:text=use%20state%20instead.-,Pitfall,-The%20React%20tree).

`hydrateRoot` will try to recover by client rendering, removing all elements in the DOM, including `<head>`. `useInsertionEffect` runs before DOM mutations, so `document.head` is `undefined` when the effect fires.

Styled-jsx's `StyleSheet.makeStyleTag` method assumes `document.head` is not `null` and is called during the `useInsertionEffect` callback. This causes the error: `Cannot read properties of null (reading 'appendChild')` and client rendering fails, resulting in a blank page.

[Codesandbox example](https://codesandbox.io/p/sandbox/47y8qk).

My opinion:

Yes, the underlying issue is a bug in the app. Completely preventing hydration mismatch is difficult, especially when mismatches can be caused by unexpected browser behavior, extensions, and race conditions. The ability to log mismatches and fix them later is a good workflow for this as testing every case is not feasible.

The reason I think styled-jsx should change is due to it causing client rendering to fail. React will try to recover from a mismatch and styled-jsx prevents recovery. Instead of a performance regression, users experience a broken page.

I'm happy to discuss this and will open a PR "fixing" this issue. I don't expect the PR to be merged as I'm not familiar with the codebase and someone probably knows of a better way to address this.

Thanks!

PR #857, opened 2025-04-28
fix: Check if adding to registry in useInsertionEffect fails and fall back to useLayoutEffect

JeremyRH:
Addresses #856.

This will call both `useInsertionEffect` and `useLayoutEffect` and a ref is used to return early in `useLayoutEffect` if `useInsertionEffect` worked.

I tried to add tests but it seems like a fairly complete DOM implementation is needed. I didn't feel comfortable adding jsdom or happy-dom.

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -1,17 +1,18 @@
-import React from 'react'
+import React, { useLayoutEffect, useRef } from 'react'
 import { useStyleRegistry, createStyleRegistry } from './stylesheet-registry'
 import { computeId } from './lib/hash'
 
 // Opt-into the new `useInsertionEffect` API in React 18, fallback to `useLayoutEffect`.
 // https://github.com/reactwg/react-18/discussions/110
-const useInsertionEffect = React.useInsertionEffect || React.useLayoutEffect
+const useInsertionEffect = React.useInsertionEffect || useLayoutEffect
 
 const defaultRegistry =
   typeof window !== 'undefined' ? createStyleRegistry() : undefined
 export default function JSXStyle(props) {
   const registry = defaultRegistry ? defaultRegistry : useStyleRegistry()
+  const insertionEffectCalled = useRef(false)
 
-  // If `registry` does not exist, we do nothing here.
+  // `registry` might not exist while server-side rendering
   if (!registry) {
     return null
   }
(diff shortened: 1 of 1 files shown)


src/style.js
```diff
@@ -2,9 +2,46 @@ import React from 'react'
...
+      registry.remove(props)
+    }
+    // props.children can be string[], will be striped since id is identical
+  }, [props.id, String(props.dynamic)])
+
+  const addedDuringInsertionEffectRef = React.useRef(false)
+
+  // Opt-into the new `useInsertionEffect` API in React 18.
+  // https://github.com/reactwg/react-18/discussions/110
+  if (React.useInsertionEffect) {
```
huozhi, line 18: Let's avoid writing hooks in the conditional invoking way, the root cause is that there's no `head` during react client recover lead to the insertion failure. We need that condition for both useInsertionEffect and useLayoutEffect hook since it's a requirement of the registry.

And we can avoid the try catch as we don't want to hide all potential legit errors. Here's a version I wrote that you can refactor based on it.

I prefer to keep having both the hook here rather than composing them all into a `useRegistry` hook as the logic in the hook is simple, not painful to duplicate, and we can avoid the name since there's already a `useStyleRegistry`

```js
useInsertionEffect(() => {
    if (!isInsertionEffectAvailable.current) {
      isInsertionEffectAvailable.current = !!document.head
      if (!isInsertionEffectAvailable.current) {
        return
      }
    }
    registry.add(props)
    return () => {
      registry.remove(props)
    }
  },
  [props.id, String(props.dynamic)]
)

React.useLayoutEffect(() => {
    if (isInsertionEffectAvailable.current || !document.head) {
      return
    }
    registry.add(props)
    return () => {
      registry.remove(props)
    }
  },
  [props.id, String(props.dynamic)]
)
```

JeremyRH: Thanks, good points. One thing I'm a little confused about is:

> avoid writing hooks in the conditional invoking way

The `if (React.useInsertionEffect) {` is used to detect if `useInsertionEffect` is available (only React 18+). This should be a stable check that never changes. Maybe a React polyfill or monkey-patch could break it.

I'll call it unconditionally in this PR.

huozhi (approved):
lgtm! thanks. CI is failing, going to address it here #858

PR #861, opened 2026-05-14
ci: declare minimum permissions on PR title workflow

arpitjain099:
Declares `permissions: contents: read, pull-requests: read` at the workflow level. The job's single step runs a PR-title validator (`amannn/action-semantic-pull-request` or equivalent) which reads the PR title via the pulls API and reports the result as the workflow's own check status. That needs read on `pull-requests` and nothing else.

The reason to be explicit even when the inherited default may already be reasonable is CVE-2025-30066, the March 2025 `tj-actions/changed-files` compromise where a tampered third-party action exfiltrated `GITHUB_TOKEN` from workflow logs. The leaked token carried whatever scope was issued at the workflow level, so per-workflow caps bound the runtime blast radius regardless of repo or org default. The block also gives drift protection if that default ever widens and is what OpenSSF Scorecard's Token-Permissions check looks for.

YAML validated locally with `yaml.safe_load`.

### DIFF
--- a/.github/workflows/prs.yml
+++ b/.github/workflows/prs.yml
@@ -6,6 +6,9 @@ on:
       - edited
       - synchronize
 
+permissions:
+  pull-requests: read
+
 jobs:
   main:
     runs-on: ubuntu-latest


styfle (approved):

ijjk (approved):

arpitjain099:
@ijjk / @styfle - PR has approval and is green. Any chance of a merge when convenient?

.github/workflows/prs.yml
```diff
@@ -6,6 +6,10 @@ on:
       - edited
       - synchronize
 
+permissions:
+  contents: read
```
styfle:
```suggestion
```

The read content permission is not necessary

styfle:
The read content permission is not necessary. Also you need to sign your commits before we can merge.

arpitjain099:
Thanks @styfle, both addressed:

- Dropped `contents: read` (you're right, the title check only reads the PR, so `pull-requests: read` is all it needs). Kept your change and credited you on the commit.
- Re-signed: the branch is now a single SSH-signed, DCO signed-off commit, so it shows as Verified.

Rebased onto current `main` while I was at it. Should be ready to merge whenever convenient.

styfle (approved):

